PR #2, opened 2020-02-01
Project 4

Jenniferot:
Link to our deployed project:
https://hungry-meninsky-cd1922.netlify.com

It turned out very well finally.

We would like to have suggestions on how we can make the js code more simple and maybe shorter.

Best from Natalie, Lisa and Jennifer/ The Tigers

### DIFF
--- a/code/pizzabot.js
+++ b/code/pizzabot.js
@@ -1,5 +1,5 @@
 // Iteration 1
-
+/*
 const vegetarian = "Vegetarian Pizza"
 const hawaiian = "Hawaiian Pizza"
 const pepperoni = "Pepperoni Pizza"
@@ -20,7 +20,7 @@ alert(`Hey, happy to serve your pizza! On our menu we have ${vegetarian}, ${hawa
 
 const orderName = prompt("Enter the name of the pizza you want to order today!");
 
-const orderQuantity = prompt(`How many of ${orderName} do you want?`);
+var orderQuantity = prompt(`How many of ${orderName} do you want?`);
 
 console.log(orderQuantity);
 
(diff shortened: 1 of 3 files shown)


ericaglimsholt (changes requested):
Hi you all! It seems like the Netlify link maybe isn't up to date? There's no way to make the integration for the page. See the screenshot

image

ericaglimsholt:
> Hi you all! It seems like the Netlify link maybe isn't up to date? There's no way to make the integration for the page. See the screenshot
> 
> image

Nevermind! `prompt` doesn't work in a child link, just a parent!

code/pizzabot.js
```diff
@@ -1,7 +1,185 @@
+// Iteration 1
```
ericaglimsholt, line 1: The code that we don't use we can delete! 😀

code/pizzabot.js
```diff
@@ -1,7 +1,185 @@
...
+const checkOrderName = (orderName) => {
+  if (orderName == hawaiian || orderName == vegetarian || orderName == pepperoni) {
+    return true;
+  } else {
+    return false;
+  }
+}
+
+//Invoking the function
+const test = checkOrderName(orderName);
```
ericaglimsholt, line 148: Maybe this variable could have a better name? That explains more what will happen if you are using it

code/pizzabot.js
```diff
@@ -1,7 +1,185 @@
...
+  } else if (orderQuantity < 6) {
+    return (15);
+  } else {
+    return (20);
+  }
+}
+
+
+const checkOrder = (test) => {
+  if (test === true) {
```
ericaglimsholt, line 170: This actually means `true`! If you want to check if something is `false` it would be `(!test)`
```suggestion
  if (test) {
```

code/pizzabot.js
```diff
@@ -1,7 +1,185 @@
...
+    orderQuantity = prompt(`How many of ${orderName} do you want?`);
+    //Invoking the function
+    orderTotal = totalCost(orderQuantity);
+    //Invoking the function
+    minutes = cookingTime(orderQuantity);
+    alert
```
ericaglimsholt, line 176: How this is on a separate row?

ericaglimsholt (approved):
Nice the hear that the collobration between you two worked good! And I hope that you found that pair programming is a such a good way of learning together. Of course you can pair on the other assignments as well! 😀

PR #4, opened 2020-02-03
Pizza bot: Emelie & Johanna

jastrand:
Link to my deployed project: https://determined-keller-320070.netlify.com/

My reflections on how this week's project turned out: The first pair programming was challenging and fun. We manage to get all the functions to work just fine, but would have loved to spend more time to design the page as well. 

Things I'd like to have clarified or explained in more detail: It would be great to have a full code along that actually creates a html form and adding javascript to it as well, that is pretty basic but still manages multiple functions and if statements.


ericaglimsholt (changes requested):
Hi Emelie and Johanna! It seems like the Netlify link doesn't work all the way, there's no integration. Maybe you have forgotten to update the deploy with new stuff?

image

ericaglimsholt:
> Hi Emelie and Johanna! It seems like the Netlify link doesn't work all the way, there's no integration. Maybe you have forgotten to update the deploy with new stuff?
> 
> image

Nevermind - `prompt` doesn't work as a link only as a parent link. Didn't know!

code/pizzabot.js
```diff
@@ -1,7 +1,55 @@
...
+  } else if (orderQuantity < 6) {
+    return (15);
+  } else {
+    return (20);
+  }
+}
+
+
+const checkOrder = (order) => {
+  if (order === true) {
```
ericaglimsholt, line 42: This means `true` actually! And `!order` means false! 
```suggestion
  if (order) {
```

ericaglimsholt (approved):
Hi!
Good job on the Javascript. I like that you have added comments to the functions - that is great. The code as well is very easy to read. Well done! A nice way of developing this more would be to add some styling to the page. Maybe output the orders in the markup?

PR #8, opened 2020-02-04
Pizza bot by Elin Aarenstrup and Emma Prené 

emmapre:
Link to my deployed project: https://pensive-khorana-971d25.netlify.com/

My reflections on how this week's project turned out: It was fun with pairprogramming and we managed to solve our problems by thinking together or asking our team for help. We're happy with the result and how clean the code looked in the end. 

Things I'd like to have clarified or explained in more detail: Since we didn't do any of the stretch goals we understand what we did qutie well. If there is any way to do anything smarter with what we know so far it would be interesting to know.

### DIFF
--- a/code/pizzabot.js
+++ b/code/pizzabot.js
@@ -34,7 +34,7 @@ alert(`Hey! Happy to serve your pizza. On our menu we have ${vegetarian}, ${hawa
 
 const orderName = prompt(`Enter the name of the pizza you want to order today.`)
 
-if (checkOrderName(orderName) === true) {
+if (checkOrderName(orderName)) {
   orderQuantity = prompt(`How many of ${orderName} do you want?`)
   document.getElementById("pizza-order").innerHTML = `Great, I'll get started on your ${orderName} right away, it will cost ${orderTotal(orderQuantity)} kr. The pizzas will take ${pizzaTime(orderQuantity)} minutes.`
 } else {
--- a/code/pizzabot.js
+++ b/code/pizzabot.js
@@ -11,3 +11,7 @@ alert(`Hey! Happy to serve your pizza. On our menu we have ${vegetarian}, ${hawa
 
 const orderName = prompt(`Enter the name of the pizza you want to order today.`)
 const orderQuantity = prompt(`How many of ${orderName} do you want?`)
+
+const orderTotal = orderQuantity * pizzaPrice
+
+alert(`Great, I'll get started on your ${orderName} right away, it will cost ${orderTotal} kr`)
--- a/code/pizzabot.js
+++ b/code/pizzabot.js
@@ -8,3 +8,6 @@ const pizzaPrice = 80
 console.log(vegetarian, hawaiian, pepperoni, pizzaPrice)
 
 alert(`Hey! Happy to serve your pizza. On our menu we have ${vegetarian}, ${hawaiian} and ${pepperoni}`)
+
+const orderName = prompt(`Enter the name of the pizza you want to order today.`)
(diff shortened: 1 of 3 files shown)


code/pizzabot.js
```diff
@@ -5,3 +5,38 @@ const pepperoni = "Pepperoni Pizza"
...
+    orderName === vegetarian || orderName === hawaiian || orderName === pepperoni
+  ) {
+    return true;
+  } else {
+    return false;
+  }
+}
+
+let orderTotal = (orderQuantity) => {
+  return +orderQuantity * pizzaPrice
```
ericaglimsholt, line 20: How come you are setting `+` before `orderQuantity` here?

code/pizzabot.js
```diff
@@ -5,3 +5,38 @@ const pepperoni = "Pepperoni Pizza"
 const pizzaPrice = 80
 
 //Put your Javscript code here:
+
+let checkOrderName = (orderName) => {
```
ericaglimsholt, line 9: You can set your functions so be `const` instead of `let` because we don't change the `checkOrderName` value!

code/pizzabot.js
```diff
@@ -5,3 +5,38 @@ const pepperoni = "Pepperoni Pizza"
...
+  } else {
+    return false;
+  }
+}
+
+let orderTotal = (orderQuantity) => {
+  return +orderQuantity * pizzaPrice
+}
+
+let pizzaTime = (orderQuantity) => {
```
ericaglimsholt:
```suggestion
const pizzaTime = (orderQuantity) => {
```

code/pizzabot.js
```diff
@@ -5,3 +5,38 @@ const pepperoni = "Pepperoni Pizza"
...
+  } else if (+orderQuantity <= 5) {
+    return 15
+  } else {
+    return 20
+  }
+}
+
+alert(`Hey! Happy to serve your pizza. On our menu we have ${vegetarian}, ${hawaiian} and ${pepperoni}`)
+
+let orderName = prompt(`Enter the name of the pizza you want to order today.`)
```
ericaglimsholt:
```suggestion
const orderName = prompt(`Enter the name of the pizza you want to order today.`)
```

code/pizzabot.js
```diff
@@ -5,3 +5,38 @@ const pepperoni = "Pepperoni Pizza"
...
+
+if (checkOrderName(orderName) === true) {
```
ericaglimsholt: This means `true` as well! If you want to check `false` you can type `!checkOrderName(orderName)`!
```suggestion
if (checkOrderName(orderName)) {
```

ericaglimsholt (approved):
Hi Elin and Emma!
I gave you some pointers for the JavaScript part but I think that you did great on this assignment! Everything works as it should be. My comments is only there to develop the code even more 😀

code/pizzabot.js
emmapre: To make it from a string (the input from the prompt) to a number. It showed undefined at first. But maybe it is already done earlier in the code so it would not be needed this time?

code/pizzabot.js
emmapre: Thanks! That's much more clear, I think I used something similar in the cluedo projects without the === true.

code/pizzabot.js
ericaglimsholt: It's needed if you have `orderQuantity` as a string before (this converts it to a integer), so that's is probably the reason! Then it's good to have this

PR #10, opened 2020-02-05
Pizza Bot - Bea Lundevall & Camilla Gausland

bealun:
If we had more time on this project, the next step would be to include pictures of the pizzas that the user can choose from, instead of text input. Also, it would be to make a timer with count down.

To sum this up, we are very proud of our pizza bot. We managed to make the functions work, and also did some of the stretch goals. We are also proud that we figured out how to send the users that met the if requirement to the final message, and how to inform a user who didn't, to try ordering again.

Deployed site: https://hardcore-swirles-3a3762.netlify.com


code/index.html
```diff
@@ -1,12 +1,38 @@
...
+ <head>
+   <meta charset="UTF-8" />
+   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+   <link rel="shortcut icon" href="img/favicon.png" type="image/x-icon">
+   <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+   <link href="https://fonts.googleapis.com/css?family=Pacifico&display=swap" rel="stylesheet">
+   <link rel="stylesheet" href="style.css">
+   <title>PIZZA BOT</title>
+   <link href="https://fonts.googleapis.com/css?family=Playfair+Display&display=swap" rel="stylesheet">
+   <link href="https://fonts.googleapis.com/css?family=Abril+Fatface|Playfair+Display&display=swap" rel="stylesheet">
```
ericaglimsholt, line 12: Tip! On Google fonts you can choose multiple fonts so you only get one `<link>` to insert here instead of two

code/pizzabot.js
```diff
@@ -1,7 +1,40 @@
-const vegetarian = "Vegetarian Pizza"
-const hawaiian = "Hawaiian Pizza"
-const pepperoni = "Pepperoni Pizza"
+const A = "Vegetarian Pizza"
```
ericaglimsholt, line 1: A better naming here would be nice! The names should explain what the data/value is. `A` doesn't say so much. It can be anything
```suggestion
const vegetarian = "Vegetarian Pizza"
```

code/pizzabot.js
```diff
@@ -1,7 +1,40 @@
...
+const B = "Hawaiian Pizza"
+const C = "Pepperoni Pizza"
```
ericaglimsholt, line 3: In other parts in this file you are using `'` instead of `"` so it would be great if your file follow the same rules!
```suggestion
const C = 'Pepperoni Pizza'
```

code/pizzabot.js
```diff
@@ -1,7 +1,40 @@
...
+amount()
+
+
+const checkOrderName = () => {
+  orderInput = document.getElementById("pizza").value
+  numberPizza = document.getElementById("numberPizza").value
+  let totalprice = pizzaPrice * numberPizza
+  if (orderInput === A || orderInput === B || orderInput === C) {
+    if (numberPizza <= 2) {
+      time = '10 min'
```
ericaglimsholt, line 27: Time is a number, which is a int. So you can instead add "min" to the line instead `${time} min` since it is something that needs to be added to every deliver time!
```suggestion
      time = 10
```

code/pizzabot.js
```diff
@@ -1,7 +1,40 @@
...
+    } else if (numberPizza >= 3 && numberPizza <= 5) {
+      time = '15 min'
+    } else {
+      time = '20 min'
+    }
+    let final = `You have ordered ${numberPizza} ${orderInput} and it will cost ${totalprice} SEK
```
ericaglimsholt, line 33:
```suggestion
    const final = `You have ordered ${numberPizza} ${orderInput} and it will cost ${totalprice} SEK 
```

code/style.css
```diff
@@ -0,0 +1,86 @@
+input {
+  border: 2px solid black;
+  background-color: rgb(238, 238, 238); 
+}
+
+/* input:focus {
+  border: 2px white;
+  box-shadow: none;
+} */
```
ericaglimsholt, line 9: We can delete code we don't use

code/style.css
```diff
@@ -0,0 +1,86 @@
...
+  margin: auto;
+  height: 90%;
+  padding: 0px 20px;
+  font-family: 'Playfair Display', serif;
+}
+
+h1{
+  text-align: center;
+  font-size: 50px;
+  /* font-family: 'Pacifico', cursive; */
```
ericaglimsholt, line 37:
```suggestion
```

ericaglimsholt (approved):
Great work on your project!
I love that you have taken the time to style the page. It looks sooo good! Good job 🤩

PR #13, opened 2020-02-05
Frida Eriksson & Christina Persson

Cilpersson:
**Link to our deployed project:**
https://worldsbestpizzabot.netlify.com/

**Our reflections on how this week's project turned out:**
 
### What We struggled with and should improve for next pair programming project
We both feel as if we should have had a timer and take small breaks through out the day. We took longer breaks for meals and such, but an additional smaller breaks would have been good. 

For next time this is something both of us will suggest to whoever our partner will be. 

### What We're proud of
TEAMWORK, we complemented each other well. We had som issues with the form at first but we managed to figure it out and be able to save the input from the user on the web page. 

Christina is super proud over Frida and the fact that she did the CSS by herself during the evening, even tho we said that we where gonna call it a day!

Frida is super proud over Christina and that she took time to explain the javascript parts in a pedagogical way, having peeked at the instructions beforehand!

## If We had more time
We would have done some cleaning in the CSS and would have liked to try to toggle so the second label and input was visible first after entering what kind of pizza the customer wanted. 

**Things We'd like to have clarified or explained in more detail:**
We tried to put the <input> inside of the <label> but when we did so the input boxes didn't show in the browser. Damien said in a video that we should do it like so for the vision impaired. Why didn't it work?


code/pizzabot.js
```diff
@@ -1,7 +1,66 @@
...
+const pepperoni = "Pepperoni Pizza";
+const menu = [vegetarian, hawaiian, pepperoni];
 
-const pizzaPrice = 80
+const pizzaPrice = 80;
 
 //Put your Javscript code here:
+
+const checkOrderName = orderName => {
+  for (let i = 0; i < menu.length; i++) {
```
ericaglimsholt, line 11: Nice that you are using a loop here!

ericaglimsholt (approved):
YOU KNOW WHAT
There's nothing to comment about. You did such a great work. I really hope that you are proud of what you have created here. I also really enjoy to read the description of this pull request!

PR #14, opened 2020-02-05
Project Pizza Bot – Kajsa Bootin & Kajsa Holmström

crushoftheyear:
Link to my deployed project:
https://suspicious-poitras-14bb4e.netlify.com/

We are happy and satisfied with the result of our Pizza Bot. We made it work without too much issues and we had a good teamwork.

### DIFF
--- a/code/index.html
+++ b/code/index.html
@@ -2,8 +2,6 @@
 <html lang="en">
 
 <head>
-
-
   <meta charset="UTF-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
   <meta http-equiv="X-UA-Compatible" content="ie=edge" />
@@ -12,8 +10,10 @@
 
 <body>
   <h1>This is your Pizza Bot</h1>
-  <h2 id="test"></h2>
+  <h2 id="message"></h2>
+  <button id="orderButton" onclick="clicked()">Order</button>
+  <script src="pizzabot.js"></script>
 </body>
-<script src="pizzabot.js"></script>
+
 
 </html>
--- a/code/index.html
+++ b/code/index.html
@@ -12,6 +12,7 @@
 
 <body>
   <h1>This is your Pizza Bot</h1>
+  <h2 id="test"></h2>
 </body>
 <script src="pizzabot.js"></script>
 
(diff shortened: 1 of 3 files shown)


code/index.html
```diff
@@ -1,12 +1,19 @@
...
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+  <title>Project 4: Pizzabot</title>
+</head>
+
+<body>
+  <h1>This is your Pizza Bot</h1>
+  <h2 id="message"></h2>
+  <button id="orderButton" onclick="clicked()">Order</button>
```
ericaglimsholt, line 14: It would be good if the function name would be more describing of what is does. `clicked()` doesn't say so much of what will happen in that function

code/pizzabot.js
```diff
@@ -1,7 +1,42 @@
...
 
-//Put your Javscript code here:
+const checkOrderName = (pizza) => {
+  for (let i = 0; i < pizzaMenu.length; i++) {
+    if (pizzaMenu[i].toLowerCase() === pizza.toLowerCase()) {
+      return true;
+    }
+  }
+  return false;
+};
```
ericaglimsholt, line 16: Nice!!!

code/pizzabot.js
```diff
@@ -1,7 +1,42 @@
...
+
+const totalCost = (quantity) => {
+  return pizzaPrice * +quantity;
+};
+
+const cookingTime = (quantity) => {
+  if (+quantity <= 2) {
```
ericaglimsholt, line 23: How come you are adding `+` before?

ericaglimsholt (approved):
Great work on your project!
I hope that you like to pair together in this assignment, and that you will help each other out on the rest of the assignments as well 😀

code/pizzabot.js
crushoftheyear: To convert string to a number. We previously used `parseFloat` or `parseInt` but then we learned from last weeks codealong that we could use the `+`operator  instead.

ericaglimsholt: That's true! But the `quantity` is never a string as I can see? :)

PR #23, opened 2020-09-10
Project Pizza Bot - Vanessa Sue Smith and Liliana Reger Horvath

VanessaSue27:
Deployed project: https://optimistic-northcutt-67f18c.netlify.app/

We worked really well as a team and managed to follow together what is going on with the code.
Since this is our first time deep diving into JavaScript, we learned about having a good overview of all the elements playing a role in our code, and the connections between them. Need to keep a close eye on what links to what.
This project was a lot about using the correct syntax, control flow and allowing for all these new concepts to set in.

A big aha! moment was to find out about the preventDefault() event method. We had started by submitting to HTTPBin in order for the code to fully run, but then learned about this method from Matilda on Stack Overflow, later on also during Wednesday's code session. We implemented it, and this was a better solution to submitting on HTTP Bin 😁
It's also very interesting to start connecting JavaScript to visual HTML elements, so it starts to become clearer what JavaScript is used for when doing Frontend.

We look forward to learning more! Lili and Vanessa ✌


code/index.html
```diff
@@ -4,9 +4,49 @@
...
+
+        <label>
+          <br>Which Pizza would you like to order?<br>
+          <input id="pizza-type" type="text" name="pizza-type" required 
+          placeholder="Vegetarian | Hawaiian | Pepperoni" />
+        </label>
+
+        <label>
+          <br>How many Pizzas would you like?<br>
+          <input id="pizza-number" type="number" name="pizza-number" min="1" max="10" required  />
```
puj: Good job with the validation here!

code/index.html
```diff
@@ -4,9 +4,49 @@
...
+
+        <label>
+          Full Name:<br>
+          <input id="userName" type="text" name="name" required />
+        </label>
+
+        <label>
+          <br>Which Pizza would you like to order?<br>
+          <input id="pizza-type" type="text" name="pizza-type" required 
+          placeholder="Vegetarian | Hawaiian | Pepperoni" />
```
puj: How do you tackle the problem of a small `placeholder` if you have, for instance, 12 pizzas?

code/pizzabot.js
```diff
@@ -4,4 +4,68 @@ const pepperoni = "Pepperoni Pizza"
...
+const submitForm = document.getElementById('form-action');
+
+alert(`Hey! Happy to serve your pizza. On our menu we have ${vegetarian}, ${hawaiian} and ${pepperoni}`);
+
+submitForm.onsubmit = submitOrder;
+
+function submitOrder(event) {
+event.preventDefault();  //Prevents the page from being refreshes when the form is submitted
+let customerName = document.getElementById('userName').value; //gets user's name
+let orderName = document.getElementById('pizza-type').value; //gets user input on type of pizza
```
puj: The entire body of this function should be indented one step.  This will become increasingly important as we have more functions in our code 👍

code/pizzabot.js
```diff
@@ -4,4 +4,68 @@ const pepperoni = "Pepperoni Pizza"
 
 const pizzaPrice = 80
 
-//Put your Javscript code here:
+const submitForm = document.getElementById('form-action');
+
+alert(`Hey! Happy to serve your pizza. On our menu we have ${vegetarian}, ${hawaiian} and ${pepperoni}`);
+
+submitForm.onsubmit = submitOrder;
```
puj, line 11: This is also a good way of doing this.  A more javascript-oriented way. We'll see more of this type of approach moving forward. 👍

code/pizzabot.js
```diff
@@ -4,4 +4,68 @@ const pepperoni = "Pepperoni Pizza"
...
+if (confirmName(orderName)) { //Calls confirmName function to check if pizza type entered by user is correct
+    let orderQuantity = document.getElementById('pizza-number').value;  //gets user input on amount of pizzas
+
+    let orderTotal = totalCost(orderQuantity); //Calls totalCost function to calculate total order cost
+
+    let orderTime = prepTime(orderQuantity); //Calls prepTime function to calculate total preparation time
+
+    let confirmText = document.getElementById('confirm');
+    let pizzaPhoto = document.getElementById('pizza-photo');
+    confirmText.style.display = "block";
```
puj: Nearly all of your variables could be `const` instead of `let`.  Generally, it's extremely good practice to use `const` **until** you need to use `let`

```javascript
    const confirmText = document.getElementById('confirm');
    confirmText.style.display = "block";
```

This is still valid, because you are changing something inside the variable, not the variable itself.

code/pizzabot.js
```diff
@@ -4,4 +4,68 @@ const pepperoni = "Pepperoni Pizza"
...
+
+    confirmText.innerHTML = `Great ${customerName}, I'll get started on your ${orderName} Pizza(s) right away, it will cost ${orderTotal} kr. The pizza(s) will take ${orderTime} minutes.`;
+    pizzaPhoto.src = pizzaImage(orderName); // Calls pizzaImage function to show the ordered Pizza's image by replacing src property
+
+} else {
+    alert('Please select a Pizza from the menu.');
```
puj: I wonder, if there was some way you could use the `onChange` event on the `pizza-type` input in order to validate the pizza name...

code/pizzabot.js
```diff
@@ -4,4 +4,68 @@ const pepperoni = "Pepperoni Pizza"
...
+    }
+};
+
+//Function to check if user's input pizza choice is correct - toLowerCase in case the user has entered with caps
+let confirmName = (pizzaName) => {
+    if (pizzaName.toLowerCase() === "hawaiian" || pizzaName.toLowerCase() ==="pepperoni" || pizzaName.toLowerCase() === "vegetarian") {
+        return true;
+    } else {
+        return false;
+    }
```
puj, line 43: Interestingly enough, you could just return the boolean.
```javascript
return pizzaName.toLowerCase() === "hawaiian" ||
      pizzaName.toLowerCase() ==="pepperoni" || 
      pizzaName.toLowerCase() === "vegetarian";
```

Another way, to be more explicit, would be:
```javascript
// Create boolean value
const isValidPizzaType = pizzaName.toLowerCase() === "hawaiian" ||
      pizzaName.toLowerCase() ==="pepperoni" || 
      pizzaName.toLowerCase() === "vegetarian";
return isValidPizzaType;
```

puj: Good job checking for cases like "HaWaIIAn" 👍 
I would prefer to see you using your variables here though.
`pizzaName.toLowerCase() === pepperoni`

code/pizzabot.js
```diff
@@ -4,4 +4,68 @@ const pepperoni = "Pepperoni Pizza"
...
+
+//Function to caltulate Total preparation time
+let prepTime = (amount) => {
+    if(amount <= 2) {
+        return 10
+    } else if (amount >= 3 && amount <= 5) {
+        return 15
+    } else {
+        return 20
+    }
```
puj: There are always a ton of ways to structure logic like this.  Here's another alternative:
```javascript
    if(amount <= 2) {
        return 10;
    } else if (amount >= 6 ) {
        return 20;
    } else {
        // Values in between
        return 15;
    }
```

code/pizzabot.js
```diff
@@ -4,4 +4,68 @@ const pepperoni = "Pepperoni Pizza"
...
+
+//Function to choose which Pizza imagee
+const pizzaImage = (pizza) => {
+    if(pizza.toLowerCase() === 'vegetarian') {
+        return "./vegetarian.jpg"
+    } else if (pizza.toLowerCase() === 'hawaiian') {
+        return "./hawaiian.jpg"
+    } else {
+        return "./pepperoni.jpg"
+    }
```
puj: Really nicely done. 
Here's an example of using a switch statement to achieve the same results.  Switch statements can usually be found when checking discrete values (like names, or exact numbers) rather than ranges (like amounts/prices).  Both are fine.
```javascript
switch(pizza.toLowerCase()) {
  case "vegetarian":
    return "./vegetarian.jpg"
    break;
  case 'hawaiian':
    return "./hawaiian.jpg"
    break;
  default:
    return "./pepperoni.jpg"
}
```

code/style.css
```diff
@@ -0,0 +1,86 @@
...
+    color: #87bf2a;
+}
+
+.one {
+    margin-bottom: 0;
+}
+
+.two {
+    margin-top: 5px;
+}
```
puj, line 41: I suppose you could avoid this by having some element wrapped around these two

puj (changes requested):
Hey Vanessa and Liliana!
Very impressive project!  
- Extremely good description of the project in your README and nice additions in your Pull Request message, keep this up!
- Really good commit messages.  Descriptive and concise!
- Your reflections on: "we might come up with a specific solution, which seems long and complex, later to find that there is a simpler and shorter way to do it"  are extremely valuable.  Continue with this mindset, there will be plenty of opportunities to put it to good use!
- Your CSS was simple and too the point, but impactful.  Nice!

Things to fix: 
- Indentation
- let -> const
- Some places you use semi-colons, some not.  It's important to be consistent on this one

puj (approved):
Nice fixes! Well done, approved!

PR #29, opened 2020-09-11
Nathalie & Destiny Pizza Bot

nathalienordstrom:
Link to my deployed project:


code/index.html
```diff
@@ -4,9 +4,147 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <link rel="stylesheet" type="text/css" href="style.css" />
+    <link
+      href="https://fonts.googleapis.com/css?family=Montserrat:400,700&display=swap"
+      rel="stylesheet"
```
EmOlsz: Very nice code formatting :)

code/index.html
```diff
@@ -4,9 +4,147 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <link rel="stylesheet" type="text/css" href="style.css" />
```

code/index.html
```diff
@@ -4,9 +4,147 @@
...
+    <link
+      href="https://fonts.googleapis.com/css?family=Montserrat:400,700&display=swap"
+      rel="stylesheet"
+    />
     <title>Project 4: Pizzabot</title>
   </head>
+
+  <div class="topnav">
+    <img class="logga" src="https://images.vexels.com/media/users/3/128435/isolated/preview/3bba8cf18c7b31717c98231e7d6e736e-italian-pizza-logo-by-vexels.png" alt="logo" ></img>
```
EmOlsz: `img` is self closing tag - it should look like this :

``

instead of :

`</img>`

code/index.html
```diff
@@ -4,9 +4,147 @@
...
+
+    <div class="container">
+      <div class="pizzabot">
+        <!-- Welcome -->
+        <div id="first"></div>
+
+        <!-- Choose pizza -->
+        <div id="second">
+          <h3>Choose your favorite pizza</h3>
+          <section class="three_pizzas">
```
EmOlsz: When naming classes, we should use `kebab-case` convention, not `snake_case` as you did, so it should look like `three-pizzas` rather than `three_pizzas`

code/index.html
```diff
@@ -4,9 +4,147 @@
...
+            <label for="vegetarian">
```
EmOlsz: If you nest `input` inside `label`, there is no need for `for="vegetarian"` attribute. We can either choose 

```
<label>
    That's my label
    <input />
</label>
```
or
```
<label for="vegetarian">That's my label</label>
<input id="vegetarian" />
```

code/index.html
```diff
@@ -4,9 +4,147 @@
...
+              <input
+                type="radio"
+                id="vegetarian"
+                name="pizzaChoice"
+                value="Vegetarian Pizza"
+              />
+              <figure>
+                <img src="vegetarian.png" alt="Vegetarian Pizza" />
```
EmOlsz: `./`

code/index.html
```diff
@@ -4,9 +4,147 @@
...
+                </figcaption>
+              </figure>
+            </label>
+          </section>
+        </div>
+
+        <!-- How many do you want  -->
+        <div id="third">
+          <h3>How many of <span id="pizzaChoice"></span> would you like?</h3>
+          <section class="pizzas_number">
```
EmOlsz, line 86: Hmm... Don't you think that `<input type="radio" />` are not the best idea over here? What if I would like to order 11 pizzas? Not mentioning about how long your code is because of that.
Far better approach would be to create one `<input type="number" />`, where user can type amount he/she wants to order

code/index.html
```diff
@@ -4,9 +4,147 @@
...
+          <button id="next_btn">Next</button>
+        </div>
+      </div>
+    </div>
+    <footer>
+      <p>&copy; Destiny & Nathalie </p>
+      <p></p>
+    </footer>
+
+    <script src="pizzabot.js"></script>
```
EmOlsz, line 147: `./`

code/pizzabot.js
```diff
@@ -1,7 +1,115 @@
+// functions
+const gotoNext = () => {
```
EmOlsz, line 9: Regarding content of this function - it's not the best idea to control behaviour of elements like that. HTML in general should be autonomous code, connected to JS only in crucial parts.
Moreover, if you want to disable any button after something happens (in this case, when `gotoNext()` or `gotoPrevious()` functions are called), we don't have to cover all of buttons and sections, but only the one which should change.

Nonetheless, this is far too complicated code, please take a look on my general comment above regarding that

code/pizzabot.js
```diff
@@ -1,7 +1,115 @@
...
+      document.getElementById("fourth").style.display = "none"
+      document.getElementById("first").style.display = "block"
+      document.getElementById("next_btn").style.display = "inline-block"
+      document.getElementById("back_btn").innerHTML = "Back"
+      document.getElementById("back_btn").style.display = "none"
+  }
+}
+
+const getOrderName = () => {
+  let orderName = document.querySelector('input[name="pizzaChoice"]:checked').value
```
EmOlsz: `getOrderName()` function is triggered each time you click on one of radio buttons. Why not use `event` argument provided by JavaScript and check value of clicked button by `event.target.value`?
Alternatively you can just pass your custom argument there, let's say `getOrderName('vegetarian')` and then, in JS, just assign this value to `orderName` variable. It could look like this :

HTML:
`<input type="radio" value="vegetarian" onclick="getOrderName(event)" />`
or
`<input type="radio" value="vegetarian" onclick="getOrderName('vegetarian')" />`

JS:
```
const getOrderName = (event) => {
  let orderName = event.target.value;
}
```
or
```
const getOrderName = (pizzaType) => {
  let orderName = pizzaType;
}
```

code/pizzabot.js
```diff
@@ -1,7 +1,115 @@
...
+  document.getElementById("pizzaChoice").innerHTML = orderName
+  document.getElementById("next_btn").disabled = false
+}
+
+const getConfirmation = () => {
+  let orderName = document.querySelector('input[name="pizzaChoice"]:checked').value
```
EmOlsz: Once again you are creating `orderName` variable. Why not creating it one time in global scope so you can access it from every function?

code/pizzabot.js
```diff
@@ -1,7 +1,115 @@
...
+  let orderQuantity = document.querySelector('input[name="numberOrdered"]:checked').value
```
EmOlsz: Same with orderQuantity - this variable could be created in global scope and being overwritten only one time - after we got exact value selected by user

code/pizzabot.js
```diff
@@ -1,7 +1,115 @@
...
+  document.getElementById("next_btn").disabled = false
+  let totalPrice = orderQuantity * pizzaPrice
```
EmOlsz: You are using a lot of `let` keyword, even though you never overwrite their values. You should use `const` instead.
General rule is that at the beginning all fo your variables should be created with `const` keyword. You should change them to `let` only after you find out you need to overwrite value of your variable somewhere in code

code/pizzabot.js
```diff
@@ -1,7 +1,115 @@
...
+      return 20
+  } else {
+      return 15
+  }
+}
+
+// constants
 const vegetarian = "Vegetarian Pizza"
 const hawaiian = "Hawaiian Pizza"
 const pepperoni = "Pepperoni Pizza"
```
EmOlsz: All of those should be at the top of JS file

code/pizzabot.js
```diff
@@ -1,7 +1,115 @@
...
+document.getElementById("btn1").addEventListener("click", getConfirmation)
+document.getElementById("btn2").addEventListener("click", getConfirmation)
+document.getElementById("btn3").addEventListener("click", getConfirmation)
+document.getElementById("btn4").addEventListener("click", getConfirmation)
+document.getElementById("btn5").addEventListener("click", getConfirmation)
+document.getElementById("btn6").addEventListener("click", getConfirmation)
+document.getElementById("btn7").addEventListener("click", getConfirmation)
+document.getElementById("btn8").addEventListener("click", getConfirmation)
+document.getElementById("btn9").addEventListener("click", getConfirmation)
+document.getElementById("btn10").addEventListener("click", getConfirmation)
```
EmOlsz: Once again - imagine doing same logic with one `<input type="number" />` instead of 10x `<input type="radio" />`

code/style.css
```diff
@@ -0,0 +1,266 @@
+.topnav {
+  color: #242424;
+  padding: 14px 16px;
+  font-size: 30px;
+  font-family: Impact, Haettenschweiler, 'Arial Narrow Bold', sans-serif;
+display: flex;
+justify-content: space-evenly;
```
EmOlsz: wrong indentation

code/style.css
```diff
@@ -0,0 +1,266 @@
...
+}
+
+/* unvisited link */
+a:link {
+    color: black;
+    text-decoration: none;
+  }
```
EmOlsz, line 14: from here below you have wrong indents

code/style.css
```diff
@@ -0,0 +1,266 @@
...
+
+
+.logga {
+  height: 200px;
+  width: 200px;
+}
+
+body {
+    background-image: url(marble-2398946_1280.jpg);
```

`background-image: url("marble-2398946_1280.jpg");`

code/style.css
```diff
@@ -0,0 +1,266 @@
...
+    font-family: Impact, Haettenschweiler, 'Arial Narrow Bold', sans-serif;
+    font-size: 20px;
+    background-repeat: repeat;
+    background-size: auto;
+    margin: 0px;
```
EmOlsz: If value is equal to `0px`, you can omit `px` and just type `0`

code/style.css
```diff
@@ -0,0 +1,266 @@
...
+    height: 800px;
+    width: 100%;
+    min-width: 300px;
+  }
+  
+
+  p {
```
EmOlsz: Why styling by selectors? We should avoid that and focus mostly on styling by class name

code/style.css
```diff
@@ -0,0 +1,266 @@
...
+  
+  footer {
+    padding: 20px;
+    display: block;
+    background-color: hotpink;
+    height: 10vh;
+    color:white;
+  }
+  
+  /* */
```
EmOlsz: Some unused comment, let's delete it

code/style.css
```diff
@@ -0,0 +1,266 @@
...
+    position: relative;
+    box-shadow: 10px 10px 5px 5px #000000;
+  }
+  
+  section.three_pizzas {
+    margin-top: 30px;
+  }
+  
+  
+  #first,
```
EmOlsz: Definitely we should avoid styling by id as well

code/style.css
```diff
@@ -0,0 +1,266 @@
...
+  
+  /* Controls */
+  .controls {
+    height: 20%;
+  
+  }
+  
+  .controls button {
+    background-image:url(marble-2398946_1280.jpg);
+    margin: 0px 0 0 0;
```
EmOlsz: we can write this line of code by shorthand - `margin: 0;`

code/style.css
```diff
@@ -0,0 +1,266 @@
...
+    width: 100px;
+    height: 60px;
+    font-size: 20px;
+    border: 1px #000000 solid;
+    border-radius: 10px;
+    text-decoration: none;
+
+
```
EmOlsz: Let's delete these empty lines

code/style.css
```diff
@@ -0,0 +1,266 @@
...
+  }
+
+  }
+  
+
+  
+  
+  
+  .numBtn:hover {
```
EmOlsz: We should only use `kebab-case` naming pattern rather than `camelCase` when it comes to class names

EmOlsz (changes requested):
You chose pretty ambitious approach Nathalie & Destiny! Building "slideshow app" mechanism is very hard even to make it work, not to mention clean, semantic way of writing it. Nonetheless, you made it work and even added some interesting changes like pizza-as-radio-button, html visualisation and, lastly, slideshow effect.

Still, there is a lot of messy code and some rather random ways of writing your logic. Please, check out my comments below and fix as much of code as you can. Before I will be able to approve your project, I need to see part of those changes, which are :

1. Changing `img` tag from `</img>` to ``
2. Changing `let` to `const`
3. Fixing indentation, commented code and empty lines

code/index.html
nathalienordstrom: do I need a . before the /?

code/index.html
nathalienordstrom: why do I have to write That's my label?

code/index.html
nathalienordstrom: I have never seen ./ ? why do I have to write a dot there?

code/index.html
nathalienordstrom: Is it a okey way of writing it if we would like to have an maximum of ten pizzas, so we don't need to change the hole styling regarding that?

code/index.html
nathalienordstrom: i don´t know where to put it?

code/pizzabot.js
nathalienordstrom: I don´t really follow what is the different outcome of is. What is the final diffrence?

code/pizzabot.js
nathalienordstrom: I can't figur out how it would look?

code/index.html
EmOlsz: Yes, `./` sign means basically `THIS folder`. As I mentioned that during the lecture, from pragmatic point of view there is no difference between paths `style.css` and `./style.css`, but from semantic and clean-code point of view we should always include `./`

code/index.html
EmOlsz: You don't have to write exactly this sentence :) What I meant is that you need to write something, no matter what.

code/index.html
EmOlsz: Check out my comment above. I talked about that during lectures, please check recordings from Week 2

code/index.html
EmOlsz: If you want to have maximum and minimum amounts, you can do that by validating your numeric input as well :

`<input type="number" min="1" max="10" />`

code/index.html
EmOlsz: Check comment above

code/pizzabot.js
EmOlsz: There is not pragmatic difference, both yours and mine code work the same. The difference lays on HOW we write it - your approach is non-semantical

code/pizzabot.js
EmOlsz: It would look like this :

`let orderName;`.                                                                                                                       // This is line no 1 of JS file
`orderName = document.querySelector('input[name="pizzaChoice"]:checked').value`      // This is line no 47

code/pizzabot.js
nathalienordstrom: But can I have the layout that is with one row of code as you typed?

EmOlsz: Yes, of course

code/index.html
nathalienordstrom: Will the text come up somehwere or what is the purpose of the text?

code/index.html
nathalienordstrom: <div id="third">
          <h3>How many of <span id="pizzaChoice"></span> would you like?</h3>
          <section class="pizzas_number">
            <label for="btn1">
         **so like this**  **<input type="number" min="1" max="10" />**
**Insted of this?** **<input type="radio" id="btn1" name="numberOrdered" value="1" ./>**
              <div class="numBtn">1</div>
            </label>

nathalienordstrom: so I just replace this line:
 <input type="radio" id="btn1" name="numberOrdered" value="1" ./>

with
<input type="number" min="1" max="10" ./>

or do i deleta everything below line 91

code/index.html
EmOlsz: Yes, that's main purpose of `label` - it informs user what the control next to it is about

code/index.html
EmOlsz: Yes, you need to delete all 10 radio buttons and replace them with only one numeric input

code/index.html
nathalienordstrom: I tried to write 
<label for="vegetarian">Vegetarian</label>
<input id="vegetarian" />

But then the word should up in ha weird place and I don´t want a word there. so can I just skip to have word there?

nathalienordstrom: I tried to write the first code example you gave and the text that I wrote shows up and I don´t want a text there. So I just put 
```
<label> 
<input
type="radio"
                id="vegetarian"
                name="pizzaChoice"
                value="Vegetarian Pizza"
              />
</label> 
```

nathalienordstrom: is that okey?

EmOlsz (approved):

PR #30, opened 2020-09-11
Ellen Björkmans and Anna Hasselströms Pizza Bot

annahasselstrom:
Link to our deployed project:
https://elegant-minsky-dce503.netlify.app/

This has been an interesting and fun yet difficult introduction to javascript. It was a challenge and although we didn't get everything to work the way we wanted, we are still proud of what we managed to do. The knowledge sharing has been super important - both between us as a pair but also within the Hippo-group. Very, very valuable.

It would have been great with a better introduction to onclick events in the beginning of the week and to have clearer instructions on how to visualise the form in the html in iteration nr 7.

- Onclick events - we would like to understand better the connection between html and js.
- Variables: when should they be declared outside or inside functions, that is understanding global vs local scope.
- Connecting js to html.


ClaireCaudwell:
I agree! An intro into onclick events would have been great! Looks great Anna and Ellen.

code/index.html
```diff
@@ -4,9 +4,64 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <link type="text/css" rel="stylesheet" href="style.css">
```
EmOlsz, line 7: Let's not forget about `./` in path to `style.css`

code/index.html
```diff
@@ -4,9 +4,64 @@
...
     <title>Project 4: Pizzabot</title>
   </head>
   <body>
-    <h1>This is your Pizza Bot</h1>
+    <section class="main-container">
+        <img src="./img/pizza1.jpg" alt="image of pizza">
+    <!--container for the form-->
+    <div class="form-container">
```
EmOlsz, line 14: Some wrong indentation - `.form-container` element is nested inside `.main-container`, but still has the same distance to left side as its parent

code/index.html
```diff
@@ -4,9 +4,64 @@
...
+      <div class="content-container">
+        <div class="content-text">
+          <h1>Pizza Time!</h1>
+          <p id="welcomeMessage"></p> 
+        </div>
+    <!--form that gets user input, id:s connects with javascript and assigns value to variables.-->
+        <form id="form-action">      
+          <fieldset>
+            <legend>Order here</legend>
```
EmOlsz, line 23: Very nice! :)

code/index.html
```diff
@@ -4,9 +4,64 @@
...
+            <div class="form-action-group">
+              <div class="choosePizza">
```
EmOlsz, line 25: why `camelCase` instead of `kebab-case` over here?

code/index.html
```diff
@@ -4,9 +4,64 @@
...
+          <h1>Pizza Time!</h1>
+          <p id="welcomeMessage"></p> 
+        </div>
+    <!--form that gets user input, id:s connects with javascript and assigns value to variables.-->
+        <form id="form-action">      
+          <fieldset>
+            <legend>Order here</legend>
+            <div class="form-action-group">
+              <div class="choosePizza"> 
+                <p id="choosePizzaText"></p>
```
EmOlsz, line 26: same here and everywhere below

code/index.html
```diff
@@ -4,9 +4,64 @@
...
+              </div>
+            
+              <label for="pizzaSort"></label>
+                <input 
+                type="text"
+                name="pizzaSort"
+                id="pizzaSort"
+                placeholder="e.g. Hawaiian Pizza"
+                />
```
EmOlsz, line 35: Happy to see proper code formatting, but all attributes should be two spaces more to the right than their opening tag

code/pizzabot.js
```diff
@@ -1,7 +1,73 @@
...
+const submitForm = document.getElementById("form-action");
+
+document.getElementById("welcomeMessage").innerHTML=(`Hey! Happy to serve your pizza. On our menu we have ${vegetarian}, ${hawaiian} and ${pepperoni}.`);
+
+/*A function that checks if userinput-choice is on menu. Returns true or false. The argument 
+is the input from the form in html and is passed as a value to the parameter "chosenPizza" in the js function. 
+Also uses the toLowerCase method to allow for users using capitals and lower case. Stores the true
+value in orderName. Is checked in if-statement: if true a particular code is run. If false, another code is run. ????*/ 
+const validateOrderName = (chosenPizza) => {
+   if (chosenPizza.toLowerCase() === vegetarian.toLowerCase() || chosenPizza.toLowerCase() === pepperoni.toLowerCase() || chosenPizza.toLowerCase() === hawaiian.toLowerCase()) {
```
EmOlsz, line 16: Great validation! I'm even more happy that you apply method `.toLowerCase()` both to variable and parameter passed from user. It's not so obvious, someone else would add 6 conditionals - 3x per lowerCase and 3x per "normal case". In your approach, you already do `toLowerCase()` on both stored variable and user text, so we don't have to worry about differences.

You can add one more thing to achieve superior clean code so to speak :

create local variables (so, inside `validateOrderName ` function) where you will store `vegetarian.toLowerCase()` and two others. Thanks to that, inside if statement you will not have to use `toLowerCase()` method and whole line will be shorter. That's clean code requirement. From mechanic point of view, you couldn't do it better that it already is :)

code/pizzabot.js
```diff
@@ -1,7 +1,73 @@
...
+      return true;
+   } else {
+      return false;
+   }
+};
+
+//Function that provides us a totalprice of order by calculating price and number of pizzas. Will run of if-statement is true.
+const calculateTotalCost = (pizzaPrice, orderQuantity) => {
+   orderTotal = (pizzaPrice * orderQuantity);
+   return orderTotal;
```
EmOlsz, line 26: No need to parentheses in `(pizzaPrice * orderQuantity)`. For mathematical operations we should use them in same cases as in mathematics from school, so only when we want to change order of calculations (+ or - in connection with * or/ ).
I would simplify this code in other way as well : content of you function can just be :
`return pizzaPrice * orderQuantity`.
And since it's only one line of content, we can simplify it even more :
const calculateTotalCost = (pizzaPrice, orderQuantity) => pizzaPrice * orderQuantity;

code/pizzabot.js
```diff
@@ -1,7 +1,73 @@
...
+//Function that provdies the total cooking time depending on number of pizzas. Will run if if-statement is true.
+const calculateCookingTime = (quantity) => {
+   if (quantity <= 2) {
+      return 10; 
+   } else if (quantity > 2 && quantity <= 5) {
+      return 15;
+   } else {
+      return 20;
+   }
+};
```
EmOlsz, line 38: Great job :) I forgot to mention one thing - if we pass only one parameter to our function, instead o f writing it like this : 

`const calculateCookingTime = (quantity) => { ... }`

We can omit parentheses wrapping our `quantity` parameter and write :

`const calculateCookingTime = quantity => { ... }`

It's once again clean code approach. You just have to remember that we can apply this technique ONLY if we have one argument

code/pizzabot.js
```diff
@@ -1,7 +1,73 @@
...
+
+document.getElementById("choosePizzaText").innerHTML=("What pizza do you want to order today?");
+   
+//Can not solve this issue. Method asking for user input. Connecting the above getElementById with this (orderName). Not working since orderName is defined inside of function. 
+//document.getElementById("amountsOfPizza").innerHTML=(`How many of ${orderName} do you want?`);
```
EmOlsz, line 43: Great job with comments! They help me a lot.
I'm not sure though what is the problem. Can you describe it in more detailed way?

code/pizzabot.js
```diff
@@ -1,7 +1,73 @@
...
+
+//Function that runs code when user clicks submit. Connects the values from the html to javascript and outputs in browser.
+const submitOrder= (event) => {
+   //Prevents the page from being refreshed when the form is submitted.
+   event.preventDefault();
```
EmOlsz, line 48: 👍

code/index.html
```diff
@@ -4,9 +4,64 @@
...
+    <!--form that gets user input, id:s connects with javascript and assigns value to variables.-->
+        <form id="form-action">      
+          <fieldset>
+            <legend>Order here</legend>
+            <div class="form-action-group">
+              <div class="choosePizza"> 
+                <p id="choosePizzaText"></p>
+              </div>
+            
+              <label for="pizzaSort"></label>
```
EmOlsz, line 29: If you don't put any text inside `label`, what's the reason for having it?

code/pizzabot.js
```diff
@@ -1,7 +1,73 @@
...
+//Function that runs code when user clicks submit. Connects the values from the html to javascript and outputs in browser.
+const submitOrder= (event) => {
+   //Prevents the page from being refreshed when the form is submitted.
+   event.preventDefault(); 
+   
+   let orderName = document.getElementById("pizzaSort").value;
+
+   /*Stores a true or false value in a variable called pizzaIsOnMenu. The value will now be possible to access outside 
+   the function validateOrderName when running the if-statement below. orderName refers to user input from form. That value is passed to the function validateOrderName. The function checks true or false. That value is then stored in the variable let isPizzaOnMenu.*/
+   let pizzaIsOnMenu = validateOrderName(orderName);
```
EmOlsz, line 54: You use a lot of `let` variables which are never overwritten. All of them should be `const`.
Apart from that, you used your `validateOrderName` in a great way!
One thing to change - by convention, variables which store boolean value (`true` or `false`) should be names with `is` prefix, so correct variable name would be `isPizzaOnMenu` rather than `pizzaIsOnMenu`

code/pizzabot.js
```diff
@@ -1,7 +1,73 @@
...
+
+   //If-statement that runs one code if true, and another if false. Uses declared functions that are stored in new variables.
+   if (pizzaIsOnMenu === true) {
```
EmOlsz, line 57: You can simplify this if statement to 

`if (pizzaIsOnMenu) { ... }`

Because your variable stores boolean value. To see that, let's just put actual values from variable `pizzaIsOnMenu` into our if statement :

(This is your code):
``if (true === true) { ... }``

(This is my code):
``if (true) { ... }``

Makes more sense now?

code/style.css
```diff
@@ -0,0 +1,142 @@
+html,
+body {
+  margin: 0 auto;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  background: white;
+  color: whitesmoke;
+  font-family: "futura", sans-serif;
```
EmOlsz, line 10: Happy to see font fallback!

code/style.css
```diff
@@ -0,0 +1,142 @@
...
+  border-radius: 6px;
+}
+
+.form-action-group {
+  margin: 20px;
+}
+
+.form-action-group input {
+  width: 100%;
+  height: 30px;
```
EmOlsz, line 58: It's better to stretch our input fields by `font-size` (it will increase height) and `padding`. Try that out!

code/style.css
```diff
@@ -0,0 +1,142 @@
...
+.submit-button {
+  display: flex;
+  justify-content: center;
+  color: whitesmoke;
+  background-color: #775256;
+  margin-top: 20px;
+  font-family: "futura", sans-serif;
+  font-size: 16px;
+  width: 70px;
+  height: 30px;
```
EmOlsz, line 70: Same as above

code/style.css
```diff
@@ -0,0 +1,142 @@
...
+  text-align: center;
+  margin: 70px;
+  font-size: 20px;
+}
+
+.emoji {
+  font-size: 70px;
+  text-align: center;
+  animation: shake .5s;
+  animation-iteration-count: infinite;
```
EmOlsz, line 84: Nice :)

EmOlsz (approved):
Great job Ellen and Anna! You've done visual implementation of this project with nice form (properly used what's most important). You HTML and CSS are very nice - can you see now why we put so much of work to learn it?
But I have to say there is one thing that makes me happy far more than nice design and proper mechanism of your project - it's the way you wrote it. Functions in you JS file are done exactly the way they should be! By that, I mean that they are defined in global scope, they have parameters passed into and they always returns something. Later on you use them in local scope of `submitOrder` function, but it's great that they are not defined there.
With that, I want to say you have understood the most important thing from Week 4, great job :)

PR #31, opened 2020-09-11
Project 4: Pizza Bot (Therése & Lillemor)

Thessan:
Link to my deployed project: https://lillemor-tessan-pizzabot.netlify.app

Project made by Therése Larsson & Lillemor Eriksson


code/index.html
```diff
@@ -8,5 +8,11 @@
   </head>
   <body>
     <h1>This is your Pizza Bot</h1>
+
+    <p id="pizzaConfirmation"></p>
```
EmOlsz: For id's and classes we should rather use `kebab-case` than `camelCase` names, so
`pizza-confirmation` rather than `pizzaConfirmation`

code/index.html
```diff
@@ -8,5 +8,11 @@
...
+    <p id="pizzaCookingTime"></p>
+
+
+  <script src="pizzabot.js"></script>
```
EmOlsz: Let's not forget about `./` in path to your JS file

code/pizzabot.js
```diff
@@ -1,7 +1,62 @@
...
+const vegetarian = "Vegetarian Pizza";
+const hawaiian = "Hawaiian Pizza";
+const pepperoni = "Pepperoni Pizza";
+const pizzaPrice = 80;
 
 //Put your Javscript code here:
+
+const orderPizza = () => {
+  alert(
+    `Hey! Happy to serve your pizza. On our menu we have ${vegetarian}, ${hawaiian} and ${pepperoni}.`
```
EmOlsz: Nice implementation of template strings :)

code/pizzabot.js
```diff
@@ -1,7 +1,62 @@
...
+  );
+  const orderName = prompt(
+    "Enter the name of the pizza you want to order today."
```
EmOlsz, line 15: it's not so long line, everything can be places in one with ` const orderName = prompt(` and ` );`

code/pizzabot.js
```diff
@@ -1,7 +1,62 @@
...
+  );
+
+  const validateOrderName = () => {
+    if (
+      orderName === vegetarian ||
+      orderName === hawaiian ||
+      orderName === pepperoni
```
EmOlsz: Well done :) There are two things I would like to mention :

1. Have you considered what will happen if user types proper pizza name, but lowercase?
So rather `vegetarian pizza` than `Vegetarian Pizza`?
Right now, your code will not accept that answer. We can easly fix that, there is a nice trick for that : you can use method `.toLowerCase()` on both `orderName` and `vegetarian`, `hawaiian` and `pepperoni` - thanks to that we will be comparing values with only lower case letters. It would look something like that :

`if (orderName.toLowerCase() === vegetarian.toLowerCase() || ... )`.

To create even more clean code, we store these values inside local variables so :

`const orderNameLower = orderName.toLowerCase();`
`const vegetarianLower = vegetarian.toLoweCase();`
`...`

So in our if statement we can write :

`if (orderNameLower === vegetarianLower || ... )`.

This is clean code practice, please fix that.

2. In your `validateOrderName` function you are using global variable `orderName`. We should pass it into our function as an argument rather than reaching for it from function. This is another clean code practice. You just have to remember that if you declare you `validateOrderName` function with argument, you need to pass it in line 33, where you call your function. Fix that as well

code/pizzabot.js
```diff
@@ -1,7 +1,62 @@
...
+    } else {
+      alert("Choose a pizza that's on the menu");
+
+      return false;
+    }
+    /*if orderName false, should prompt again, or loop from start,
+     */
+  };
+
+  if (validateOrderName(orderName)) {
```
EmOlsz: Above if statement we can create new variable in which we will store result of `validateOrderName` function. Thanks to that, inside if statement we don't have to call our function, it will be called before and pass it's result into variable. Later on, we can use this variable in if statement.
There is a convention of how we should write boolean variables - they should always start with prefix 'is'. In our case, our variable could have a name `isOrderNameValid` or something like that.

Nevertheless, I'm happy you omitted `=== true` and just wrote 

` if (validateOrderName(orderName)) { ... }` 

rather than 

` if (validateOrderName(orderName) === true) { ... }`

Both these lines are the same, but first one is more clean :)

code/pizzabot.js
```diff
@@ -1,7 +1,62 @@
...
+    const orderQuantity = prompt(`How many ${orderName}s do you want?`);
+    //then prompts how many pizzas you want
+    const orderTotal = orderQuantity * pizzaPrice;
```
EmOlsz: This line of code should be stored in separate function and called in specific moment. You should fix that

code/pizzabot.js
```diff
@@ -1,7 +1,62 @@
...
+      alert("Choose a pizza that's on the menu");
+
+      return false;
+    }
+    /*if orderName false, should prompt again, or loop from start,
+     */
+  };
+
+  if (validateOrderName(orderName)) {
+    const orderQuantity = prompt(`How many ${orderName}s do you want?`);
```
EmOlsz: In this prompt box I wrote '-1'. You owe me 80kr for my pizza!
(Reproduce my steps and check out by your own ;) )

EmOlsz (changes requested):
Good job Therese and Lillemor! Your logic flow is proper, in general project works as it should. You did pizza name validation and redirect user to previous step if something is wrong. So, basic requirements are almost done, congratulations :)

Almost, because I would like to fix couple of things :

1. Implement `toLowerCase()` logic I mention in comments
2. Change `validateOrderName` function as mentioned in comments
3. Validate values lower than 0 when it comes to amount of pizza as mentioned in comments
4. Most important - you haven't met requirement of putting your code into functions. By that requirement we didn't mean you should wrap everything inside one big function `orderPizza`, but to split parts of your code which represents each step of application flow

EmOlsz (approved):

PR #32, opened 2020-09-11
Peggy & Sofia's InstaPizza bot

pwangy:
**Link to my deployed project:**
https://pwangy-pizzabot.netlify.app/

Sofia Vaz Sousa - I’m really glad and proud of what myself and Peggy accomplished in this project, we had a vision for it in terms of functionality and we did it, even though, it was a really tough week for me regarding JavaScript. I had a hard time getting my head around few concepts and then working in pairs with all of those new concepts was challenging but I know it will get better with time and experience.  :slightly_smiling_face:

Peggy Wang - I'm pleased with what we're handing in this week. It was definitely challenging but I very much appreciated having a partner to talk through whatever problems we were currently solving for and being able to put our heads together. I am sure it would have taken twice as long without her! I know there's a ton more JavaScript to be acquired, but I feel like this first deep dive was a giant step in getting there. 


blue_level/index.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <link rel="stylesheet" href="style.css" />
```
EmOlsz: Let's not forget about `./` in path to style file

blue_level/index.html
```diff
@@ -0,0 +1,27 @@
...
+        <img src="./robotchef.png" alt="robotchefimage" />
+        <p id="final-order"></p>
+      </div>
+      <div class="button" id="button" onclick="order()">
+        <p>Order pizza in just 3 steps.</p>
+        <p class="call-to-action">Click me to get started!</p>
+      </div>
+    </section>
+
+    <script src="pizzabot.js"></script>
```
EmOlsz: `./` over here as well

blue_level/pizzabot.js
```diff
@@ -0,0 +1,54 @@
+const vegetarian = "Vegetarian Pizza";
+const hawaiian = "Hawaiian Pizza";
+const pepperoni = "Pepperoni Pizza";
+const pizzaPrice = 80;
+
+const validateOrderName = (orderName) => {
```
EmOlsz: There is an arrow function convention which says that if we pass to our function only one argument, we can omit parentheses `()`, and rather write :

`const validateOrderName = orderName => { ... }`

than

`const validateOrderName = (orderName) => { ... }`

blue_level/pizzabot.js
```diff
@@ -0,0 +1,54 @@
...
+  if (
+    orderName === "vegetarian" ||
+    orderName === "pepperoni" ||
+    orderName === "hawaiian"
+  ) {
+    return true;
+  } else {
+    return false;
+  }
```
EmOlsz: Great job :)
Apart from returning given thing from function, `return` keyword does one more thing - it automatically finishes function. That's why we mustn't write anything below `return`, because that code will never be executed (of course in your case we have code below `return true`, but it's inside if statement, so it wouldn't enter second statement in the first place).

```
const validateOrderName = (orderName) => {
  if (orderName === "vegetarian" || orderName === "pepperoni" || orderName === "hawaiian") {
    return true;
  }
  return false;
}
```

As you can see, we are not writing `else` statement anymore. We can do that thanks to behaviour mentioned above; if result of if statement is `true`, we will execute `return true` code from line 12 and finish function. If, on the other hand, result of it is `false`, we will never enter line 12 and jump straight into line 14. After executing line 14 we will finish function same way :)

blue_level/pizzabot.js
```diff
@@ -0,0 +1,54 @@
...
+};
+
+const calculateTotalCost = (orderQty, pizzaPrice) => {
+  return orderQty * pizzaPrice;
+};
```
EmOlsz: Very nice :) Another tip from my side - since we have only one line inside `calculateTotalCost` function, we can write it in "arrow function shorthand" :

`const calculateTotalCost = (orderQty, pizzaPrice) => orderQty * pizzaPrice;`

blue_level/pizzabot.js
```diff
@@ -0,0 +1,54 @@
...
+
+const calculateCookingTime = (orderQty) => {
```
EmOlsz: `const calculateCookingTime = orderQty => { ... }`

blue_level/pizzabot.js
```diff
@@ -0,0 +1,54 @@
...
+    return 20;
+  }
+};
+
+const order = () => {
+  const orderName = prompt("Enter the name of the pizza you want to order today.");
+  if (validateOrderName(orderName)) {
+    let orderQty = prompt(`How many of ${orderName} do you want?`);
+    let orderTotal = calculateTotalCost(orderQty, pizzaPrice);
+    let pizzaCookingTime = calculateCookingTime(orderQty);
```
EmOlsz: Why `let`? They are never being overwritten. Let's change it to `const`

blue_level/pizzabot.js
```diff
@@ -0,0 +1,54 @@
...
+    document.getElementById("final-order").innerHTML = `Great! I will get started on your ${orderName} right away, it will cost ${orderTotal}kr. The pizza(s) will be ready in ${pizzaCookingTime} minutes.`;
+    document.getElementById("button").style.display = "none";
+    document.getElementById("menu").style.display = "none";
```
EmOlsz, line 48: That's great you know that there is option to manipulate styles via JS, but this antipattern - what you can do instead is to manipulate `classList` property and add or remove classes.

blue_level/style.css
```diff
@@ -0,0 +1,41 @@
+body {
+    font-family: Arial, sans-serif;
```
EmOlsz, line 2: Happy to see font-fallback :)

blue_level/style.css
```diff
@@ -0,0 +1,41 @@
...
+  }
+  
+  
+  .main-container { 
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+  }
+
+  #robotchef img {
```
EmOlsz: We shouldn't style by `id`, let's change it to `class`

index.html
```diff
@@ -0,0 +1,32 @@
...
+  </head>
+  
+  <body>
+    <h1>&#127829; InstaPizza &#127829;</h1>
+    <div class="button" id="button" onclick="buttonClicked()">
+      <p>The easiest way to order pizza.</p>
+      <p class="call-to-action">Click me to get started!</p>
+    </div>
+    
+    <div class="chat-container" id="chatContainer">
```
EmOlsz, line 18: `id`, same as `class`, should have name in `kebab-case` pattern, not `camelCase`. Don't be afraid to add same name to both `id` and `class` though, it's completely ok :)
Same thing is addressed to lines below

index.html
```diff
@@ -0,0 +1,32 @@
...
+      <div class="speech-bot bot-triangle" id="botBubble"></div>
+      <div class="speech-user user-triangle" id="userBubble" onclick="startOrder()"></div>
+    <div class="speech-bot bot-triangle" id="startOrder"></div>
+    <div class="speech-bot bot-triangle" id="confirmation"></div>
+    </div>
+
+    <footer>
+      <p><span>&#129418; </span> This page was made by Peggy @<a href="https://blipsandclicks.com">blipsandclicks</a> & Sofia Vaz Sousa for educational purposes.
+      </p>
```
EmOlsz: Wrong indentation

index.html
```diff
@@ -0,0 +1,32 @@
...
+    </footer>
+
+    <script src="pizzabot.js"></script>
```
EmOlsz: `./`

pizzabot.js
```diff
@@ -0,0 +1,83 @@
+const vegetarian = "Vegetarian Pizza";
+const hawaiian = "Hawaiian Pizza";
+const pepperoni = "Pepperoni Pizza";
+const pizzaPrice = 80;
+
+document.getElementById(
+  "botBubble"
+).innerHTML = `Hey! Happy to get your order started. On our menu we have ${vegetarian}, ${hawaiian}, and ${pepperoni}.`;
```
EmOlsz: Interesting line break, I would rather suggest :

```
document.getElementById("botBubble").innerHTML =
    `Hey! Happy to get your order started. On our menu we have ${vegetarian}, ${hawaiian}, and ${pepperoni}.`;
```

pizzabot.js
```diff
@@ -0,0 +1,83 @@
...
+
+document.getElementById("userBubble").innerHTML =
+  "Ok! (Click to start your order.)";
```
EmOlsz, line 10: Exactly as over here, haha

pizzabot.js
```diff
@@ -0,0 +1,83 @@
...
+
+document.getElementById("userBubble").style.cursor = "pointer";
```
EmOlsz: CSS in JS ⛔️
Same is addressed to lines below

pizzabot.js
```diff
@@ -0,0 +1,83 @@
...
+ 
+ <p><input type="radio" id="hawaiian" name="pizza_selection" value="hawaiian" required /><label for="hawaiian">${hawaiian}</label></p>
+      
+ <p><input type="radio" id="pepperoni" name="pizza_selection" value="pepperoni" required /><label for="pepperoni">${pepperoni}</label></p></div>
+ 
+ <div class="quantity"><p><label for="orderQty">How many pizzas?</label></p><p><input id="orderQty" type="number" name="orderQty" placeholder="enter a number" required /></p></div>
+ 
+ <p><button class=".orderButton" type="submit" id="submit" name="submit" value="submit">Order</button></p>
+ </form>`;
+}
```
EmOlsz: I know it's a lot of work, but we shouldn't write code like that, this is completely not readable - try to format this HTML so we can actually see some structure

pizzabot.js
```diff
@@ -0,0 +1,83 @@
...
+
+const calculateTotalCost = (finalOrderQty, pizzaPrice) => {
+  return finalOrderQty * pizzaPrice;
+};
+
+const calculateCookingTime = (finalOrderQty) => {
```
EmOlsz: That's already 320kr waiting for me

blue_level/pizzabot.js
```diff
@@ -0,0 +1,54 @@
...
+    return 15;
+  } else {
+    return 20;
+  }
+};
+
+const order = () => {
+  const orderName = prompt("Enter the name of the pizza you want to order today.");
+  if (validateOrderName(orderName)) {
+    let orderQty = prompt(`How many of ${orderName} do you want?`);
```
EmOlsz: I typed -2 as pizza amount, you owe me -160kr.
Please, fix that :)

EmOlsz (changes requested):
Super nice job Peggy and Sofia! I'm happy to see so much of a progress in both of yours code :) You made basic as well as advanced pizza bots and what's most important both of them are written in very clean, elegant way. That's the most important thing, believe me.
I've added some comments to your code, but consider them as a boost to you already well-maintained knowledge.

There are also some things I would like to fix :
1. Add validation to `orderQty` part - let's say if you will do that, we can forget about those 320kr ;)
2. let -> const

Those are two super small things. In general, you impressed me a lot!

blue_level/index.html
pwangy: Working on these changes now. :)

blue_level/pizzabot.js
pwangy: This is a great tip. Thank you!

blue_level/pizzabot.js
pwangy: And again! Thank you for showing us how to write cleaner code.

blue_level/pizzabot.js
pwangy: :D

pizzabot.js
pwangy: moving this styling to CSS.

pizzabot.js
pwangy: adding in formatting too!

blue_level/pizzabot.js
pwangy: changed!

blue_level/style.css
pwangy: added class to HTML and updated CSS to reflect.

index.html
pwangy: leaving this for now and will keep in mind for future projects.

pizzabot.js
pwangy: updated this.

blue_level/pizzabot.js
pwangy: For some reason VS Code keeps flagging the version without curly brackets as an error. :(

So when the code looks like this: `const calculateTotalCost = (orderQty, pizzaPrice) => return orderQty * pizzaPrice;`

pwangy: done!

pwangy:
Still working through these but did a push so that Sofia can see where I've landed in the revisions.

EmOlsz (approved):

pwangy:
WHEW! Validation successfully added to both versions of the bot so that the user cannot order 0 or even a negative number of pizzas. @EmOlsz, we no longer owe you any kr!

PR #33, opened 2020-09-11
Emelie and Tara's pizzabot!

emeliesv:
Link to my deployed project:

https://emtapizzabot.netlify.app/

We worked very well on this project together. We are happy with the way it turned out and that it works!
If we had more time we would have added a few extra features from the black level.
Proud of our communication, that we divided the work equally and did both pair-programming and programming on our own where we read each others code and talked about different solutions.


README.md
```diff
@@ -1,13 +1,19 @@
 # Pizza Bot
 
-Replace this readme with your own information about your project. 
+This is a pizza-bot-project. It's a javascript chatbot that helps you order pizza.
+It doesn't really order any pizza. Yet.
```
EmOlsz, line 4: xD

code/index.html
```diff
@@ -1,12 +1,86 @@
...
-</html>
+
+<head>
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+  <meta property=“og:title” content=“EmeTa pizza restaurant”>
+  <meta property=“og:description” content=“Best pizza in town!“>
+  <meta property=“og:image” content=“https://emtapizzabot.netlify.app/assets/hawaiian.png”>
+  <meta property=“og:url” content=“https://emtapizzabot.netlify.app/”>
```
EmOlsz, line 11: Super nice you added these meta tags!

code/index.html
```diff
@@ -1,12 +1,86 @@
...
+  <link rel="stylesheet" href="style.css">
```

code/index.html
```diff
@@ -1,12 +1,86 @@
...
+    </header>
+
+    <main>
+
+      <div class="greeting-message">
+        <form onsubmit="welcomeMessage(event)">
+          <h3>Welcome to the EmTa's pizzabot!</h3>
+          <label for="name" class="name-input">Before we start, what's your name?</label>
+          <input id="name" name="name" type="text" />
+          <input type="submit" name="submit" id="submit" value="Start pizza order">
```
EmOlsz: These two inputs have different closures - one of them has ` />` and second one `>`. Let's unify that to ` />`

code/index.html
```diff
@@ -1,12 +1,86 @@
...
+        </form>
+        <div id="messageBox">
```
EmOlsz: `id` names should be written in `kebab-case` rather than `camelCase` - let's change that

code/index.html
```diff
@@ -1,12 +1,86 @@
...
+          <p id="message" class="message-display"></p>
+        </div>
+      </div>
+
+      <div class="hidden" id="pizzaMenu">
```
EmOlsz: Same here - `kebab-case`

code/index.html
```diff
@@ -1,12 +1,86 @@
...
+        <h2>PizzaMenu</h2>
+        <p>
+        <ul>
```
EmOlsz, line 40: Oops, wrong indentation. `ul` tag is nested inside `p` tag but still it has same space amount from left side as `p`

code/index.html
```diff
@@ -1,12 +1,86 @@
...
+          <li><img src="./assets/Vegetarian.png" />Vegetarian Pizza<p class="pizzaprice">80 kr</p>
+          </li>
```
EmOlsz, line 43: 1. `img` tag should be in separate line
2. This is not the best use case for `p` tag. If you want to style only one part of a sentence/text different way than the main one, you should use `span` and add `display: block` to it so it will be displayed in next line in browser

code/index.html
```diff
@@ -1,12 +1,86 @@
...
+          <li><img src="./assets/fusion.png" />Fusion Pizza<p class="pizzaprice">120 kr</p>
+          </li>
+          <li><img src="./assets/margarita.png" />Margaritha Pizza<p class="pizzaprice">65 kr</p>
+          </li>
+          <li><img src="./assets/kebab.png" />Kebab pizza<p class="pizzaprice">85 kr</p>
+          </li>
+        </ul>
+        </p>
+        <label for="selectedPizza">Select your pizza</label>
```
EmOlsz: `kebab-case` - over here and everywhere below

code/index.html
```diff
@@ -1,12 +1,86 @@
...
+        <input type="number" name="pizzaQuantity" id="pizzaQuantity" />
+
+        <button onclick="showTotalOrder()">CLICK</button>
+
+      </div>
+      <div class="hidden" id="totalOrder">
+        <h3>Yummy!</h3>
+        <p id="finalOrderMessage"></p>
+        <img id="pizzaImg" class="finalPizza">
+        <button onclick="welcomeMessage(event); hide()">Refresh</button>
```
EmOlsz: We should avoid adding two functions into one event inside HTML. Best way would be calling `hide()` function inside `welcomeMessage(event)` function at the very bottom (or, at this case, before `welcomeMessage(event)`)

code/pizzabot.js
```diff
@@ -1,7 +1,103 @@
...
+const vegetarian = "vegetarian pizza";
+const hawaiian = "hawaiian pizza";
+const pepperoni = "pepperoni pizza";
+const margaritha = "margaritha pizza";
+const kebab = "kebab pizza";
+const fusion = "fusion pizza";
 
-const pizzaPrice = 80
+const welcomeMessage = (event) => {
+  event.preventDefault();
```
EmOlsz, line 9: Great job :) Just one thing - if we pass only one parameter to our function, instead of writing it like this :

`const welcomeMessage = (event) => { ... }`

We can omit parentheses wrapping our quantity parameter and write :

`const welcomeMessage = event => { ... }`

It's clean code approach. You just have to remember that we can apply this technique ONLY if we have one argument.

Of course same comment is relevant to other functions of yours

code/pizzabot.js
```diff
@@ -1,7 +1,103 @@
...
+  const username = document.getElementById("name").value;
+  showMessageBox(username);
+};
 
-//Put your Javscript code here:
+const showMessageBox = (username) => {
+  document.getElementById(
+    "message"
+  ).innerHTML = `Hey ${username}! Let's start our order! Select a pizza from our menu:`;
```
EmOlsz: I would rather suggest different line break, like this one :

```
  document.getElementById("message").innerHTML =
    `Hey ${username}! Let's start our order! Select a pizza from our menu:`;
```

code/pizzabot.js
```diff
@@ -1,7 +1,103 @@
...
+  showPizzaMenu();
+};
+
+const showPizzaMenu = () => {
+  document.getElementById("pizzaMenu").classList.remove("hidden");
```
EmOlsz: Great job :) especially I like that you used CSS classes rather than adding `style.display = "none"` directly.
Main reason for that is adding exact styles via JS is consider as antipattern. As convention says - only code inside CSS files should be responsible for handling styles to avoid chaos in code

code/pizzabot.js
```diff
@@ -1,7 +1,103 @@
...
+
+const validateOrderName = (selectedPizza) => {
+  let selectedPizzaLower = selectedPizza.toLowerCase();
+  return (
+    selectedPizzaLower === vegetarian ||
+    selectedPizzaLower === hawaiian ||
+    selectedPizzaLower === pepperoni ||
+    selectedPizzaLower === fusion ||
+    selectedPizzaLower === margaritha ||
+    selectedPizzaLower === kebab
```
EmOlsz, line 32: Clever, I'm not gonna lie - I checked whether this is working :D

code/pizzabot.js
```diff
@@ -1,7 +1,103 @@
...
+  );
+};
+
+const getPizzaPrice = (selectedPizza) => {
+  if ((selectedPizza = "margaritha")) {
```
EmOlsz: Three things over here :
1. You mustn't use single equation mark (`=`) in if statements. In JavaScript, `=` mark is called `Assignment mark`, and can only be used to assign a value (mostly, we do that for variables - we assign some value to our variable, both newly created or created some time ago and overwritten right now).
On the other hand `==` and `===` are `Comparison marks` used to compare values and only those should be used inside conditional statements.
Try to test your application - as you can see, no matter whether we choose hawaiian, papperoni or other pizza, price for single pizza is always 65kr because your conditionals are not working properly
2. You don't need to wrap `(selectedPizza = "margaritha")` inside another pair of parentheses, it's completely ok with one pair already :)
3. You are comparing parameter to string - why not comparing parameter to variable? Btw even if you will fix comparison marks to work, every pizza of yours will cost 80kr because neither you have `kebab` or `fusion` - you have `kebab pizza` and `fusion pizza` so until that one is fixed as well, all if statements will be `false`

code/pizzabot.js
```diff
@@ -1,7 +1,103 @@
...
+  let validation = validateOrderName(selectedPizza);
+  if (validation) {
+    document.getElementById("quantity").classList.remove("hidden");
+  } else {
+    document.getElementById("errorMessage").classList.remove("hidden");
+  }
+};
+
+const calculateCookingTime = (orderQuantity) => {
+  if (orderQuantity > 0 && orderQuantity <= 2) {
```
EmOlsz, line 66: I tried to pass value `-2` over there and it works - you owe me 130kr (well, being honest I have pretty big sum of money already after testing this week projects of students).
This one needs to be fixed or you will not be the most profitable pizzeria in town

code/pizzabot.js
```diff
@@ -1,7 +1,103 @@
...
+  } else if ((selectedPizza = "fusion")) {
+    return 120;
+  } else {
+    return 80;
+  }
+};
+
+const showQuantity = () => {
+  const selectedPizza = document.getElementById("selectedPizza").value; //add .toLowerCase?
+  let validation = validateOrderName(selectedPizza);
```
EmOlsz: Why `let`? We are not overwriting this value anywhere so it can (and should) be `const`

code/pizzabot.js
```diff
@@ -1,7 +1,103 @@
...
+  if (validation) {
```
EmOlsz, line 51: I like that you wrote `if (validation)` instead of `if (validation === true)`, great job :)

code/pizzabot.js
```diff
@@ -1,7 +1,103 @@
...
+  }
+};
+
+const showTotalOrder = () => {
+  document.getElementById("totalOrder").classList.remove("hidden");
+  const pizzaQuantity = document.getElementById("pizzaQuantity").value;
+  const cookingTime = calculateCookingTime(pizzaQuantity);
+  const selectedPizza = document
+    .getElementById("selectedPizza")
+    .value.toLowerCase();
```
EmOlsz: Wow, you made my day :) Keep on doing that!

code/pizzabot.js
```diff
@@ -1,7 +1,103 @@
...
+  const totalCost = getPizzaPrice(selectedPizza) * pizzaQuantity;
```
EmOlsz: Super small thing, but from clean code perspective it would be better to split this out into two lines :

1. `const pizzaPrice = getPizzaPrice(selectedPizza)`
2. `const totalCost = pizzaPrice * pizzaQuantity;`

code/style.css
```diff
@@ -0,0 +1,131 @@
+@import url("https://fonts.googleapis.com/css2?family=Raleway:ital,wght@0,100;0,400;0,600;0,700;0,800;1,100;1,400;1,600;1,700;1,800&display=swap");
```
EmOlsz, line 1: Great, custom font :)

code/style.css
```diff
@@ -0,0 +1,131 @@
...
+
+* {
+  box-sizing: border-box;
+  font-family: "Raleway", sans-serif;
```
EmOlsz, line 5: And font fallback!

code/style.css
```diff
@@ -0,0 +1,131 @@
...
+}
+
+.grid-container {
+  display: grid;
+  grid-template-columns: repeat(8, 1fr);
```
EmOlsz, line 10: That's a good idea to experiment with something different than flexbox

code/style.css
```diff
@@ -0,0 +1,131 @@
...
+  gap: 5px;
+}
+
+header {
```
EmOlsz: Why you haven't add class for `header` as well? We should be consequent, if we style by classes, we should style mostly or - in perfect situation - only by them

code/style.css
```diff
@@ -0,0 +1,131 @@
...
+  grid-column: span 8;
+  text-align: center;
+  background-color: #84a59d;
+  color: #fefae0;
+  font-size: 3vh;
```
EmOlsz: Hmm, it's not the best idea to set `font-size` equal to 3 viewport height units. I don't mean it's wrong or it wouldn't work, but it just doesn't make any sense. Viewport units are only used for height and width. Best units for `font-size` are `px`, `em`, `rem` and `%`

code/style.css
```diff
@@ -0,0 +1,131 @@
...
+
+.greeting-message {
+  grid-column-start: 2;
+  grid-column-end: 6;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-content: center;
+  background: #f6bd60;
+  padding: 5vh;
```
EmOlsz, line 45: Same over here - doesn't make sense to use `vh`

code/style.css
```diff
@@ -0,0 +1,131 @@
...
+}
+
+.name {
+  margin: 10px, 0px;
```
EmOlsz: This line of code is not going to work, there shouldn't be any `,`. Moreover, if value is `0px`, we can omit `px` and just write `0`

code/style.css
```diff
@@ -0,0 +1,131 @@
...
+.name-input {
+  margin: 15px;
+}
+
+.message-display {
+  text-align: center;
+}
+
+input {
+    margin: 10px;
```
EmOlsz: Wrong indentation

code/style.css
```diff
@@ -0,0 +1,131 @@
...
+  width: 30vh;
+  padding: 10px;
+}
+
+.quantity {
+  margin: 20px;
+}
+
+.finalPizza {
+    max-width: 80%
```
EmOlsz: Another wrong indentation

EmOlsz (changes requested):
Congratulations Emelie and Tara, very nice project! I'm impressed by the amount of extra requirements you've done this week :) Both custom pizzas, visualised HTML approach and images of pizza are present, moreover "chat-like" experience is there! I also like how modern your JS files is in regards of line breaking and ES6 usage, great job!
On the other hand I have strong impression of completely not equal level of attention you paid into your project. It has some bugs that could be detected simply by testing it for 10 minutes - that seem to be thing you should focus on during next weeks. Moreover, your HTML and CSS files looks messy - indents and some units should be fixed.

Apart from some comments which you can consider as knowledge-boost (your knowledge level is already pretty high, so I encourage you to check them out since it looks like you will absorb them), there are also some things you need to fix in order to have Week 4 accepted. Those are :

1. If statement inside `getPizzaPrice()` function is not working - check all three points I've added there
2. Change all `let` occurrences which are not overwritten anywhere into `const`
3. Add validation for orderQuantity value - we shouldn't be able to pass numbers lower than 1 there
4. HTML and CSS cleanup

Still, even with these issues above, you made great job, I'm impressed a lot :)

EmOlsz (approved):

PR #35, opened 2020-09-11
Project 4: Pizzabot Karin Nordkvist & Gabriella Bolin 

Gabbi-89:
Link to our deployed project: https://pizzeria-sliceandsnack-bot.netlify.app/ 
@karinnordkvist & @gabbi-89

For us it was super fun with pair-programming, even though we both were a bit uncertain beforehand. We are both very happy with the site and that it looks like a Chatbot. 

Things we would like to have clarified or explained in more detail:
We had problems with getting the last function to work according to the syntax we learnt, but with help from Maks we understood that we had tried to call the function before declaring it. So in the end it all worked.

If more time we would also like to learn about and try using objects/arrays but we assume we will get more time for that in the coming weeks.


code/index.html
```diff
@@ -4,9 +4,162 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
-    <title>Project 4: Pizzabot</title>
+    <link rel="stylesheet" type="text/css" href="./style.css"/>
+    <link rel="icon" type="image/png" href="./images/favicon.ico">
```
EmOlsz, line 8: Happy to see both favicon and proper path to files (with `./` included)

code/index.html
```diff
@@ -4,9 +4,162 @@
...
+    <title>Pizzeria SLICE & SNACK - Pizzabot</title>
   </head>
   <body>
-    <h1>This is your Pizza Bot</h1>
+  
+    <!-- ----------------------------- HEADER ----------------------------- -->
```
EmOlsz, line 13: Well commented code is crucial part of good practices - great job!

code/index.html
```diff
@@ -4,9 +4,162 @@
...
+    <header>
+      <img src="./images/pizza-slice2.png" alt="Picture of a pizzaslice, company logo">
```
EmOlsz: Let's just not forget about writing self-closing tags this way `` instead of this way ``

code/index.html
```diff
@@ -4,9 +4,162 @@
...
+      <h1>PIZZERIA SLICE & SNACK!</h1>
+    </header>
+
+   <!-- ----------------------------- MAIN ----------------------------- -->
+    <main>
+
+      <!-- ---------- PROGRESS BAR ---------- -->
+      <div id="progressBar" class="progress-bar-container-outer invisible">
+        <div class="progress-bar-container-inner">
+          <div class="progress-bar-container-inner-content">
```
EmOlsz: I see BEM methodology, it's good you remembered about it :)
Small thing to notice, these classes could be written in slightly different fashion :

`.progress-bar__container-outer` because `progress-bar` is BLOCK, and `container-outer` is ELEMENT.
Class invisible could be implemented over here as `.progress-bar__container-outer_invisible` since it's MODIFIER.

At the end your element could look like this :
```
<div 
    id="progressBar"
    class="progress-bar__container-outer progress-bar__container-outer_invisible"
></div>
```

Usage of all three of those class types creates BEM (`BLOCK__ELEMENT_MODIFIER`) and that would be 100% clean approach

code/index.html
```diff
@@ -4,9 +4,162 @@
...
+        <div class="user-inner">
+          <p>Select one of the pizza's below:</p>
+          <div class="chat-form">
+            <form>
+              <input 
+                id="pepperoni"
+                type="button"
+                value="Pepperoni Pizza"
+                onclick="orderPizza('Pepperoni Pizza')"
+                />
```
EmOlsz: Great job with line breaks :) You missed some indents though, `/>` in line 67 should be in the same vertical level as `<input` in line 62

code/index.html
```diff
@@ -4,9 +4,162 @@
...
+          <p>Select number of pizza(s):</p>
+          <div class="chat-form">
+            <form>
+              <input 
+              type="number" 
+              id="quantityUser" 
+              name="quantity" 
+              value="1"
+              min="1" 
+              max="10">
```
EmOlsz: Wrong closing tag

code/index.html
```diff
@@ -4,9 +4,162 @@
...
+             <button 
+              onclick="quantityFunction(event)"
+              >Order!</button>
```
EmOlsz: No need for line break over here, `button` tag is relatively short

code/index.html
```diff
@@ -4,9 +4,162 @@
...
+          <p id="pizza" class="pizza"></p>
+          <p id="price" class="price"></p>
+        </div>
+
+        <div class="order-total">
+          <p id="total"></p>
+        </div>
+
+        <div class="receipt">
+          <a href="#"><button>Email me a receipt</button></a>
```
EmOlsz: Since `button` is nested inside `a`, it should be placed in separate line like this :

```
<a href="#">
    <button>Email me a receipt</button>
</a>
```

The only exception to this rule are inline tags, `span` for example. Main clean-code convention says that all tags that has `display` value different than `inline`, should be written in separate line in HTML. Only those with `inline` value can occur on the same line as parent tag (`b`, `i`, `span` etc).

code/pizzabot.js
```diff
@@ -1,7 +1,138 @@
...
+// Welcome message - Pizza-chat box 2
+document.getElementById("welcome").innerHTML = `On our menu we have ${vegetarian}, ${hawaiian}, ${pepperoni} and ${kebab}.`;
+
+// Declaring variables
+let orderName; 
+
+// Makes Pizza-chat box 3 + User-chat box 2 visible. Function is called in next function.
+let displayOrderQuantityBox = () => {
+    const orderQuantityBox = document.getElementById("orderQuantityBox");
+    const selectQuantityBox = document.getElementById("selectQuantityBox");
```
EmOlsz, line 18: very well named variables!

code/pizzabot.js
```diff
@@ -1,7 +1,138 @@
...
+
+    orderQuantityBox.classList.remove("invisible");
+    orderQuantityBox.classList.add("visible");
+    selectQuantityBox.classList.remove("invisible");
+    selectQuantityBox.classList.add("visible");
```
EmOlsz, line 23: That's very good that you manipulate `classList` content of HTML tags rather than directly manipulating styles via `element.style.someStyleToManipulate = "newValue"`, because this is antipattern.
Main reason for that is, as convention says - only code inside CSS files should be responsible for handling styles to avoid chaos in code.
Great job!

code/pizzabot.js
```diff
@@ -1,7 +1,138 @@
...
+    }
+
+// Main order-function to select pizza sort + invokes function above
+let orderPizza = (name) => {
```
EmOlsz: why `let` and not `const`? I don't see any place where you overwrite value of `orderPizza` function.

Moreover, there is an arrow function convention which says that if we pass to our function only one argument, we can omit parentheses `()`, and write :

`const orderPizza = name => { ... }`

rather than

`const orderPizza = (name) => { ... }`

Important thing to notice - this convention should be applied ONLY when we pass one argument. If we would do this with multiple arguments, it would cause SyntaxError to occur

code/pizzabot.js
```diff
@@ -1,7 +1,138 @@
...
+    orderName = name;
+    document.getElementById("howMany").innerHTML = `OK, how many ${orderName}(s) do you want?`;
+    
+    // Show progress-bar when clicking on a pizza-sort
```
EmOlsz, line 31: These are good comments, they help me a lot to understand you logic flow fully, nice job :)

code/pizzabot.js
```diff
@@ -1,7 +1,138 @@
...
+    progressBar.classList.remove("invisible");
+    progressBar.classList.add("visible");
+
+    // Push header-section down when progress-bar is shown
+    const headerBox = document.querySelector("header");
+    headerBox.classList.add("marginTop");
+    
+    // Add pizza-name + first progress-box to progress-bar
+    document.getElementById("topBarPizza").innerHTML = `${orderName}`;
+    document.getElementById("progressOne").style.background = "#01DC3E";
```
EmOlsz, line 42: And over here I see one exception from `classList` approach - consider using class, maybe somethings such as `progress progress_one` where generic `progress` class is responsible for all main styles (and should be placed directly in HTML from the beginning) and modifier `progress_one` is responsible only for changing background-color

code/pizzabot.js
```diff
@@ -1,7 +1,138 @@
...
+
+    displayOrderQuantityBox()
+}
+
+// Counting-function to select amount of pizzas
+let quantityFunction = (event) => {
```
EmOlsz: once again, `let` -> `const`

code/pizzabot.js
```diff
@@ -1,7 +1,138 @@
...
+            if (orderName === kebab)
+            {
+                pizzaPrice = 100;
+            } 
+            else {
+                pizzaPrice = 80;
+            }
+        }
+        
+        otherPrice()
```
EmOlsz: You seem to add semicolon `;` from time to time but forgets to add it everywhere. Be consequent, either delete from everywhere or use it everywhere

code/pizzabot.js
```diff
@@ -1,7 +1,138 @@
...
+
+    displayOrderQuantityBox()
+}
+
+// Counting-function to select amount of pizzas
+let quantityFunction = (event) => {
+    event.preventDefault();
+
+    // Get quantity from user input-counter 
+    let orderQuantity = document.getElementById("quantityUser").value;
```
EmOlsz: I don't see any place where you would overwrite this value as well, `let` -> `const`

code/pizzabot.js
```diff
@@ -1,7 +1,138 @@
...
+    const confirmationBox = document.getElementById("confirmationBox");
+    
+    // Total cost-function
+    let orderTotal;
+    let calculateTotalCost = () => {
+        
+        // Conditional statement given other price for kebab pizza
+        let otherPrice = () => {
+            if (orderName === kebab)
+            {
```
EmOlsz: no need to put this curly bracket `{}` in next line, it can be (and even should) done like this :

```
if (orderName === kebab) {
   ...
}
```

code/style.css
```diff
@@ -0,0 +1,314 @@
+/* ---------------------- GENERAL STYLING ---------------------- */ 
+@import url('https://fonts.googleapis.com/css2?family=Roboto+Mono:wght@500&display=swap');
```
EmOlsz, line 2: Good job!

code/style.css
```diff
@@ -0,0 +1,314 @@
...
+
+* {
+    -webkit-box-sizing: border-box;
+    -moz-box-sizing: border-box;
+    box-sizing: border-box;
```
EmOlsz, line 7: Happy to see cross browser styles :)

code/style.css
```diff
@@ -0,0 +1,314 @@
...
+    -webkit-font-smoothing: antialiased;
+    text-rendering: optimizeLegibility;
+}
+
+html {
+    font-family: 'Roboto Mono', monospace;
```
EmOlsz, line 13: And font fallback!

code/style.css
```diff
@@ -0,0 +1,314 @@
...
+    margin: auto;
+    text-shadow: 1px 1px #004D16;
+}
+
+h1 {
+    font-size: 40px;
+}
+
+h2 {
+    margin: 0 auto 30px auto;
```
EmOlsz: We can use a shorthand for that : `margin: 0 auto 30px` will do the same.
Basically, CSS analyses it this way : 
1. Ok, `0` will go for `top` and `bottom` values for now, until they will tell me later on during this line that I should apply different amount for `bottom`.
2. Later, I have `auto` for `left` and `right` until they will tell me later during this line that I should apply different amount for `left`.
3. Next, I have `30px`, so that one clearly goes for `bottom` as I expected.
4. And then there is nothing, no 4th values. Hmm, so I have `0 auto 30px`. I will apply `0` for `top`, `auto` for both `left` and `right` and at the end `30px` for `bottom`

code/style.css
```diff
@@ -0,0 +1,314 @@
...
+    text-align: center;
+}
+
+p {
+    margin: 0;
+    padding: 0;
+}
+
+/* ---------------------- HEADER ----------------------*/ 
+header {
```
EmOlsz, line 41: You seem to style a lot of tags by selector name. We should avoid that, consider changing all of them to classes to keep consistency

code/style.css
```diff
@@ -0,0 +1,314 @@
...
+    margin: 60px 20px 30px 20px;
```
EmOlsz: once again, `margin: 60px 20px 30px;` is ok

code/style.css
```diff
@@ -0,0 +1,314 @@
...
+    justify-content: space-between;
+}
+
+.progress-bar {
+    width: 100%;
+    display: flex;
+}
+
+.progress-bar div {
+    height: 1vh;
```
EmOlsz, line 102: 👍

code/style.css
```diff
@@ -0,0 +1,314 @@
...
+    border-radius: 10px;
+    margin-left: 20px;
+}
+
+.user {
+    align-self: flex-end;
+}
+
+.user-inner {
+    padding: 15px 20px 15px 20px;
```
EmOlsz: shorthand `padding: 15px 20px;` is completely ok

code/style.css
```diff
@@ -0,0 +1,314 @@
...
+    margin-right: 20px;
+}
+
+/*  PORTRAITS  */ 
+.chat-portrait {
+    width: 80px;
+    height: 80px;
+    border: 2px solid #01DC3E;
+    border-radius: 50%;
+    background-image: url(./images/pizza-slice2.png);
```
EmOlsz: Some browsers will not format that path because of lack of quotes. It's far more safe to define `background-image` property like this :

`background-image: url("./images/pizza-slice2.png");`

code/style.css
```diff
@@ -0,0 +1,314 @@
...
+    align-items: center;
+}
+
+.time {
+    font-size: 50px;
+    align-self: center;
+}
+
+/*  ORDER SUMMARY  */ 
+.orderSummaryBox {
```
EmOlsz: classes should not be written in `camelCase` convention but in `kebab-case` one

code/to-do
```diff
@@ -0,0 +1,28 @@
+BACKLOG:
```
EmOlsz, line 1: Very nice backlog, happy to see that :) That tells a lot about your work maintenance. Nonetheless, we shouldn't upload this kind of files to github repo

EmOlsz (changes requested):
I'm impressed a lot ladies! This is some fine piece of project, not gonna lie :) You managed to create proper, bug-free application, you visualised it by HTML, added progress bar and chat-like conversation. This is a lot! Mostly I'm happy about how you wrote you JS logic, it's clean and well commented - keep it this way!

Nonetheless, there are couple things I would like you to work on :
1. Pay attention to indentations, you sometimes seem to forget about them
2. At the beginning, always create variable via `const` keyword. If you find out later on that you need to overwrite its value, only then change it to `let`
3. Pay attention to CSS shortcuts

Before I will approve your project, please fix indents and change `let` to `const`

Those are super minor issues, you can be 100% proud of what you have achieved this week :)

code/pizzabot.js
```diff
@@ -1,7 +1,138 @@
...
+    const countdownEl = document.getElementById("time");
+
+    const updateCountdown = () => {   
+        const minutes = Math.floor(time / 60);
+        let seconds = time % 60;
+
+        seconds = seconds < 10 ? '0' + seconds : seconds;
+    
+        countdownEl.innerHTML = `${minutes}:${seconds}`;
+        time--;
```
EmOlsz, line 129: Your timer is counting time down even after reaching 00:00 - that leads to time lower than 0 later on

Gabbi-89: @EmOlsz We are totally aware of that but we didn't get it to work unfortunately.. Do you have any tip? :)

code/style.css
karinnordkvist: We thought it was unnecessary to add classes to a semantic tag if there was only one? We only have one header, and  in the header there is only one image and one p-tag. We would definitely add a class if there were multiple variants of the same selector/element. Is this ok? Also- aren't we supposed to use the semantic tags when possible? Is it just the styling selectors we're talking about here?

code/style.css
Gabbi-89: @EmOlsz 
We see that we have had some different methods that we will change, but  to calrify what we have done wrong... 
as I understood it: In HTML we write the class names in `kebab-case` and id names in` camelCase`.
For example: <div id="summaryBox" class="summary-box"
Is that right or could you clarify a bit more? :)

code/style.css
EmOlsz: Semantic tags are there mostly to address :
1.1 HTML structure (so any other developer can easly look for `<header />` tag in your code when searching for block of code usually presented on top of the page and usually containing navigation, logo and main website settings)
2. SEO (semantically done websites are ranked with better grade by google search engines and therefore appear in google suggestions on top).

What's most important for CSS however is consistency - if we decides to use classes as styling reference, all elements should obey this rule. It's a convention mostly because main assumption done during development is "I will probably change this code later on". It doesn't really matter if you will or not, what matters are boundaries that we impose on our code. If you would decide, let's say 6 months from now on, that you want to change some styles in header, there is just big possibility you will not search for reference already created in line 41 of your CSS file but rather create new one (it's just how it is, we as humans don't think about it all the time). That would lead to some overwriting of styles. Problem is that styles applied via selector and styles applied via class name have different importance. Id's have even different importance, and styling by attributes as well. And at this point there is chaos. And we can avoid all of that by reference to classes only.

Your projects until now are pretty small so it might sound too puristic to you, but believe me, if you will get used to bad practices now, it will bite you later on in less expected moment

code/style.css
EmOlsz: As Van mentioned yesterday during lecture, naming id's by `camelCase` and `kebab-case` is almost 50/50.
Classes should be named by `kebab-case` in 100% of occurrences
Id's should be named by `kebab-case` right now, but be aware that you might find yourself looking at someone else code where id's will be done by `camelCase` and it's also OK

code/pizzabot.js
EmOlsz:
```
    const updateCountdown = () => {   
        const minutes = Math.floor(time / 60);
        let seconds = time % 60;

        seconds = seconds < 10 ? '0' + seconds : seconds;

        countdownEl.innerHTML = `${minutes}:${seconds}`;
        if (time === 0) {
                return;
                // or
                return false;
        }
        time--;
}
```

Let's try that :)

code/pizzabot.js
Gabbi-89: @EmOlsz Solved :)

EmOlsz (approved):

EmOlsz:
Great job ladies, thank you - project is officially approved :)

PR #36, opened 2020-09-11
Project 4 - Petra Almgren och Sara Stjernberg

stjernberg:
Link to my deployed project:
https://pizzeriabolero.netlify.app/

The pair program went very well. We could contribute with different things. 

We want to know how to clear radio buttons and input field (type=number).
We created a for loop for the pizza name. We would like to know how to make a similar loop for the matching pictures. We got an html collection when we tried to acces all images.  We also want to set display none to the images that should not be visible. We managed to do it with several lines, but could we do it in the loop as well?


code/index.html
```diff
@@ -1,12 +1,79 @@
...
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
-    <title>Project 4: Pizzabot</title>
-  </head>
-  <body>
-    <h1>This is your Pizza Bot</h1>
-  </body>
-</html>
+    <link rel="stylesheet" href="style.css">
```

code/index.html
```diff
@@ -1,12 +1,79 @@
...
+    <link href="https://fonts.googleapis.com/css2?family=Grandstander&family=Lato&family=Merriweather:ital,wght@0,400;1,900&display=swap" rel="stylesheet">
```
EmOlsz, line 9: Great to see custom font :)

code/index.html
```diff
@@ -1,12 +1,79 @@
...
+                     <input type="radio" name="pizza" id="chickenPizza" value="Chicken Pizza">
+                     Chicken pizza
+                    </label>
+
+                <label>
+                    <p>How many pizzas would you like?</p>
+                    <input type="number" id="quantity" name="quantity" min="1" max="10" />
+                </label>
+
+                <button onclick="receiveData(event)">Order</button>
```
EmOlsz: No need to add event `onclick` over here if you already added it to `form` tag as `onsubmit` event.
You need to add `type="submit"` attribute though

code/index.html
```diff
@@ -1,12 +1,79 @@
...
+    <main>
+
+        <section class="left-side">
+
+            <div class="welcome-box">
+                <h1>Welcome to Pizzeria Bolero</h1>
+                <p>Each pizza costs 80 kr.</p>
+                Order your pizza from the form below.
+            </div>
+            <form onsubmit="recieveData">
```
EmOlsz: You forgot about parentheses after function name. It should be written like this :

`<form onsubmit="recieveData(event)">`

rather than :

`<form onsubmit="recieveData">`

code/index.html
```diff
@@ -1,12 +1,79 @@
...
+
+            </form>
+        </section>
+
+        <section class="right-side">
+            <div class="message-container">
+                <h1>Here's your order:</h1>
+                <p id="message" class="message"> </p>
+
+                <img id="hawaiianImg" class="pizzaImg" src="./img/hawaiian.jpg" alt="hawaiian pizza">
```
EmOlsz: Self closing tags should be written this way :

`hawaiian pizza`

rather than :

`hawaiian pizza`

code/index.html
```diff
@@ -1,12 +1,79 @@
...
+                <img id="pepperoniImg" class="pizzaImg" src="./img/pepperoni.png" alt="pepperoni pizza">
+                <img id="vegetarianImg" class="pizzaImg" src="./img/vegetarisk.jfif" alt="vegetarian pizza">
+                <img id="chickenImg" class="pizzaImg" src="./img/chicken.jpg" alt="chicken pizza">
+            </div>
+        </section>
+    </main>
+
+    <footer>Created by Petra Almgren & Sara Stjernberg @Technigo September 2021</footer>
+    <script src="pizzabot.js"></script>
```
EmOlsz: Once again, `./` in file path missing

code/pizzabot.js
```diff
@@ -1,7 +1,60 @@
...
+    const typeOfPizza = validateOrderName();
+    const orderQuantity = document.getElementById("quantity").value;
+    const totalCost = calculateTotalCost(orderQuantity);
+    const cookingTime = calculateCookingTime(orderQuantity);
+    document.getElementById("message").innerHTML = (`You've ordered ${orderQuantity} ${typeOfPizza}. It will cost ${totalCost} kr and take ${cookingTime} minutes. Enjoy your meal!`);
+    showImage(typeOfPizza);
+}
 
-//Put your Javscript code here:
+let orderName;
```
EmOlsz: This variable is not used anywhere, no need to have it at all. If I'm not wrong, you use varuable `typeOfPizza` for that

code/pizzabot.js
```diff
@@ -1,7 +1,60 @@
...
+const validateOrderName = () => {
+
+    let pizzaChoices = document.getElementsByName('pizza');
+
+    for (let i = 0; i < pizzaChoices.length; i++) {
+        if (pizzaChoices[i].checked) {
+            return pizzaChoices[i].value;
```
EmOlsz, line 19: Clever :)

code/pizzabot.js
```diff
@@ -1,7 +1,60 @@
...
+        }
+    }
+}
+
+///Calculate total cost of the order.
+const calculateTotalCost = (orderQuantity) => {
+    const pizzaPrice = 80;
+    return orderQuantity * pizzaPrice;
```
EmOlsz, line 27: Why not declaring `pizzaPrice` above, at the top of the file next to pizzas?
If we would do that, we could later on simplify this code, content of your function could just be :

`return pizzaPrice * orderQuantity;`

And since it's only one line of content, we can simplify it even more :

`const calculateTotalCost = (pizzaPrice, orderQuantity) => pizzaPrice * orderQuantity;`

We can do that because in arrow functions we can omit curly brackets (`{}`) and return keyword if we only have one line of code inside.

code/pizzabot.js
```diff
@@ -1,7 +1,60 @@
...
+const calculateCookingTime = (orderQuantity) => {
+    let timeCooking;
+    if (orderQuantity < 3) {
+        timeCooking = 10;
+    } else if (orderQuantity > 5) {
+        timeCooking = 20;
+    } else {
+        timeCooking = 15;
+    }
+    return timeCooking;
```
EmOlsz, line 41: Very nice, you seem to understand how `return` keyword works :)

code/pizzabot.js
```diff
@@ -1,7 +1,60 @@
...
+
+///Calculate total cost of the order.
+const calculateTotalCost = (orderQuantity) => {
+    const pizzaPrice = 80;
+    return orderQuantity * pizzaPrice;
+}
+
+
+//Calculate cooking time
+const calculateCookingTime = (orderQuantity) => {
```
EmOlsz, line 32: Great job :) I forgot to mention one thing - if we pass only one parameter to our function, instead of writing it like this :

`const calculateCookingTime = (orderQuantity) => { ... }`

We can omit parentheses wrapping our quantity parameter and write :

`const calculateCookingTime = orderQuantity => { ... }`

It's once again clean code approach. You just have to remember that we can apply this technique ONLY if we have one argument

code/pizzabot.js
```diff
@@ -1,7 +1,60 @@
...
+        timeCooking = 20;
+    } else {
+        timeCooking = 15;
+    }
+    return timeCooking;
+}
+
+const showImage = (typeOfPizza) => {
+
+    document.getElementById("vegetarianImg").style.display = "none";
```
EmOlsz: That's great you know that there is option to manipulate styles via JS, but this antipattern - what you can do instead is to manipulate classList property and add or remove classes.

code/pizzabot.js
```diff
@@ -1,7 +1,60 @@
...
+    document.getElementById("hawaiianImg").style.display = "none";
+    document.getElementById("pepperoniImg").style.display = "none";
+    document.getElementById("chickenImg").style.display = "none";
+
+    if (typeOfPizza === "Vegetarian Pizza") {
```
EmOlsz, line 50: It would be better to store these string values inside variables - it's another convention that's there to avoid typos in code.

Imagine situation in which you have many conditionals like this (let's say 5) and in each one of them somewhere you check whether `typeOfPizza === 'Vegetarian Pizza'`. 4 out 5 times you wrote it well, but for the 5th time you forgot to add ` "Pizza"` and end up with conditional `typeOfPizza === 'Vegetarian'`. Or your pizza menu changed and you need to go over all 5 occurrences and change it there.

To avoid these kind of situations, convention says that we should pass this values into variables (in global scope, probably on top of your page), name them `vegetarianPizza` or something like that and later on use ONLY these variables everywhere. Thanks to that we are sure that all of these variable occurrences always has the same value - value that is being stored inside variable :)

code/style.css
```diff
@@ -0,0 +1,189 @@
+html {
+    position: relative;
+    min-height: 100%;
+}
+
+body {
+    background: #e88073;
+    margin: 0;
+    font-family: 'Lato', sans-serif;
```
EmOlsz, line 9: Great to see font fallback :)

code/style.css
```diff
@@ -0,0 +1,189 @@
...
+}
+
+
+/* -------header---------- */
+
+.header-image {
+    position: relative;
+    height: 300px;
+    overflow: hidden;
+    background: url(./img/headerImg.jpeg) no-repeat center;
```
EmOlsz: Some browsers will not format that path because of lack of quotes. It's far more safe to define `background` property like this :

`background: url("./img/headerImg.jpeg") no-repeat center;`

code/style.css
```diff
@@ -0,0 +1,189 @@
...
+    padding: 20px;
+    border-radius: 10px;
+    margin-top: 50px;
+}
+
+label {
+    display: block;
+}
+
+form p {
```
EmOlsz: You are not styling by class names almost at all. Styling by selectors is considered bad practice

code/style.css
```diff
@@ -0,0 +1,189 @@
...
+
+button {
+    margin-top: 20px;
+    margin-left: 50px;
+    padding: 5px;
+    background: #ba4338;
+    border-radius: 5px;
+}
+
+#quantity {
```
EmOlsz: We shouldn't also style by ids - only by class name

EmOlsz (changes requested):
Good job Petra and Sara! You've managed to write good, properly working application, moreover you visualised it nice way and even added images for each pizza element, congratulations! :) Take a look on my comments, I hope they will be helpful and will boost up your knowledge even more.

Nonetheless, there are couple of things I would like to fix before I can approve your project. Those are :
1. Please fix `onsubmit` and `onclick` method in HTML as I suggested
2. Delete unused code
3. Your application allows me to type orderQuantity as a number low than zero - this leads to bizzare situation when price for my pizza is `-160kr` - fix that as well by adding validation :)

EmOlsz (approved):

PR #38, opened 2020-09-11
Project Pizza Bot - Linda Hintze & Jonna Ru Billengren

lindahz:
**Link to my deployed project:**

https://your-pizza-bot.netlify.app/
https://github.com/lindahz/project-pizza-bot

It went well with the pair programming , better than expected. The JS was really challenging, especially creating the functions and connecting them to the form. Since there are so many ways to do it (when you google), it took time to actually understand what works and what does not. 

We started with iteration 1-7 and have kept  (commented out) some of the code from those steps to show the process of the project. 

We also completed some of the red level requirements: the form and and images that appear when submitting the form. 

We had a hard time understanding how to use JS to check what radio button is checked and what value it has, once submitting the form. We would like more examples and explanations of how to check the values from different types of html elements.


code/index.html
```diff
@@ -4,9 +4,81 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <link rel="stylesheet" href="style.css">
```
EmOlsz, line 7: Let's not forget about `./` in path to file

code/index.html
```diff
@@ -4,9 +4,81 @@
...
     <title>Project 4: Pizzabot</title>
   </head>
   <body>
-    <h1>This is your Pizza Bot</h1>
-  </body>
+
+    <header>
+
+      <video class="computer" autoplay muted loop playsinline>
+        <source src="./pizza-header.mp4" type="video/mp4">
```
EmOlsz, line 16: `video` tag implemented also this week - happy to see that :)

code/index.html
```diff
@@ -4,9 +4,81 @@
...
+        <p class="video-message">This browser does not support video...</p>
+      </video>
+
+      <h1 class="header-title">Welcome to your Pizza Bot</h1>
+
+  </header>
+
+  <main>
+    
+<div class="form-container">
```
EmOlsz, line 26: some wrong indentation

code/index.html
```diff
@@ -4,9 +4,81 @@
...
+    <form name="pizzaForm" onsubmit="handleSubmit(event)">
+
+      <div>
+        <h2 class="pizza-order">Pizza order</h2>
+      </div>
+      <div class="order">
+        <input class="input" id="userName" type="text" placeholder="Type your name!" required>
+        </label>
```
EmOlsz, line 34: Hmm, I don't see any opening `label` tag, you only have closing tag part. This needs to be fixed

code/index.html
```diff
@@ -4,9 +4,81 @@
...
+      </div>
+
+      <div>
+        <h2>Select a pizza:</h2>
```
EmOlsz, line 38: Is there really a need to wrap `h2` inside `div` over here?

code/index.html
```diff
@@ -4,9 +4,81 @@
...
+      </div>
+    
+    <div class="order-radio">
+      <label class="radio-radio">
```
EmOlsz, line 42: I admire your consequence of naming classes inside this block of code (postfix is always `-radio`), but this is not the best idea to have a class `radio-radio` ;)

code/index.html
```diff
@@ -4,9 +4,81 @@
...
+        <input class="custom-radio" id="pepperoniPizza" type="radio" name="pizza" value="Pepperoni Pizza"> Pepperoni Pizza
+      </label>
+    </div>
+
+    <p id="demo"></p>
+
+    <div class="order">
+      <label>
+        <input id="numberOfPizzas" type="number" placeholder="Select a number of pizza(s)" required>
```
EmOlsz: There is no need to add `label` if it's going to be empty - either add some text so your label is actually visible in browser or delete this wrapping `label` tag

code/index.html
```diff
@@ -4,9 +4,81 @@
...
+    <div class="order-btn">
+      <input class="btn" type="submit" value="Place Order">
+    </div>
+
+    <p id="orderSummery" class="message"></p>  
+
+    </form>
+  </div>
+
+    <div class="image-pizza">
```
EmOlsz, line 70: Once again, different indentation

code/index.html
```diff
@@ -4,9 +4,81 @@
...
+      <img src="vegetarian.jpg" id="veg" alt="vegetarian-pizza">
```
EmOlsz, line 71: `./` in file path missing

code/pizzabot.js
```diff
@@ -1,7 +1,95 @@
...
+const pizzaPrice = 80;
+
+const pizzaVegImg = document.getElementById("veg");
+const pizzaHawImg = document.getElementById("haw");
+const pizzaPepImg = document.getElementById("pep");
+
+
+// !! HIDING THIS PART SINCE IT ONLY WAS PART OF ITERATION 1-7. !!
+/*** Function 1 - Order Name 
+    const validateOrderName = (order) => {
```
EmOlsz, line 15: There is an arrow function convention which says that if we pass to our function only one argument, we can omit parentheses (), and rather write :

`const validateOrderName = order => { ... }`

than

`const validateOrderName = (order) => { ... }`

code/pizzabot.js
```diff
@@ -1,7 +1,95 @@
...
+        if (order === vegetarian || order === hawaiian || order === pepperoni) {
+          return true;
+        } else {
+          return false;
+        }
```
EmOlsz, line 20: Great job :)
Apart from returning given thing from function, return keyword does one more thing - it automatically finishes function. That's why we mustn't write anything below return, because that code will never be executed (of course in your case we have code below return true, but it's inside if statement, so it wouldn't enter second statement in the first place).
Thanks to that feature of return keyword, we can write our function in slightly different, far more clean fashion :

```
const validateOrderName = order => {
  if (orderName === "vegetarian" || orderName === "pepperoni" || orderName === "hawaiian") {
    return true;
  }
  return false;
}
```

As you can see, we are not writing else statement anymore. We can do that thanks to behaviour mentioned above; if result of if statement is `true`, we will execute `return true` code from line 17 and finish function. If, on the other hand, result of it is `false`, we will never enter line 17 and jump straight into line 19. After executing line 19 we will finish function same way :)

code/pizzabot.js
```diff
@@ -1,7 +1,95 @@
...
+      };
+***/
+
+// *** Function 2 - Total Cost
+const calculateTotalCost = (quantity, price) => {
+  return quantity * price;
+}
```
EmOlsz, line 27: Very nice :) Another tip from my side - since we have only one line inside calculateTotalCost function, we can write it in "arrow function shorthand" :

code/pizzabot.js
```diff
@@ -1,7 +1,95 @@
...
+
+// *** Function 3 - Cooking Time
+const calculateCookingTime = (time) => {
+  if (time <= 2) {
```
EmOlsz, line 31: You are not validating value lower than `0` - I was able to order -2 hawaiian pizzas, pay -160kr and wait 10 minutes for you to prepare it :) let's fix that

code/pizzabot.js
```diff
@@ -1,7 +1,95 @@
...
+const handleSubmit = (event) => {
+  event.preventDefault();
+  let userName = document.forms.pizzaForm.userName.value;
+  // makes first letter to uppercase and the rest of the mane to lowercase
+  userName = userName.charAt(0).toUpperCase() + userName.slice(1).toLowerCase();
+
+  const pizzaType = document.forms.pizzaForm.elements.pizza.value;
+  const orderQuantity = document.forms.pizzaForm.numberOfPizzas.value;
+  
+    const orderTotal = calculateTotalCost(orderQuantity, pizzaPrice);
```
EmOlsz, line 51: wrong indentation

code/pizzabot.js
```diff
@@ -1,7 +1,95 @@
...
+    const cookingTime = calculateCookingTime(orderQuantity);
+
+    // 'ternary operator' (puts the 's for plural number of pizzas)
+    const plural = orderQuantity > 1 ? "'s" : "";
```
EmOlsz, line 55: Super!!!

code/pizzabot.js
```diff
@@ -1,7 +1,95 @@
...
+
+    document.getElementById(
+      "orderSummery"
+    ).innerHTML = `Hello ${userName}! I'll get started on your ${orderQuantity} ${
+      pizzaType + plural
+    } right away. It will cost ${orderTotal} kr and will take ${cookingTime} minutes.`;
```
EmOlsz, line 61: That's very original line breaking = wouldn't it be better to format it like this?

```
document.getElementById("orderSummery").innerHTML =
    `Hello ${userName}! I'll get started on your ${orderQuantity} ${pizzaType + plural} right away. It will cost ${orderTotal} kr and will take ${cookingTime} minutes.`;
```

code/pizzabot.js
```diff
@@ -1,7 +1,95 @@
...
+    
+    if (pizzaType === "Vegetarian Pizza") {
+      pizzaVegImg.setAttribute ("style", "display:block;");
```
EmOlsz, line 64: You have access to values `"Vegetarian Pizza"` and others inside variables defined at the top of the file - let's use them instead of actual strings.

Imagine situation in which you have many conditionals like this (let's say 5) and in each one of them somewhere you check whether `pizza === 'Vegetarian Pizza'`. 4 out 5 times you wrote it well, but for the 5th time you forgot to add `' Pizza'` and end up with conditional `pizza === 'Vegetarian'`. Or your pizza menu changed and you need to go over all 5 occurrences and change it there.

To avoid these kind of situations, convention says that we should use ONLY variables everywhere. Thanks to that we are sure that all of these variable occurrences always has the same value - value that is being stored inside variable :)

code/pizzabot.js
```diff
@@ -1,7 +1,95 @@
...
+    } else if (pizzaType === "Hawaiian Pizza") {
+      pizzaHawImg.setAttribute ("style", "display:block;")
+    } else if (pizzaType === "Pepperoni Pizza") {
+      pizzaPepImg.setAttribute ("style", "display:block;")
```
EmOlsz, line 68: That's great you know that there is option to manipulate styles via JS, but this antipattern - what you can do instead is to manipulate `classList` property and add or remove classes.

code/pizzabot.js
```diff
@@ -1,7 +1,95 @@
...
+    }
+}; 
+
+/* !!  HIDING THIS PART SINCE IT ONLY WAS PART OF ITERATION 1-7 !!
+alert(
+    `Hey! Happy to serve your pizza. On our menu we have ${vegetarian}, ${hawaiian} and ${pepperoni}`
+  );
```
EmOlsz, line 75: No need for this line breaks, everything in one line is completely ok - this is not sooo long line

code/pizzabot.js
```diff
@@ -1,7 +1,95 @@
...
+
+const orderName = prompt(
+    "Enter the name of the pizza you want to order today."
+  );
+*/
+
+/* !! HIDING THIS PART SINCE IT ONLY WAS PART OF ITERATION 1-7 !!
+if (validateOrderName(orderName) === true) {
```
EmOlsz, line 83: In this case you can omit `=== true` and just write 

`if (validateOrderName(orderName)) { ... }`

Because `validateOrderName()` function itself returns `boolean` values only.
Moreover, it would be nice to store result of that function inside some variable, so our code would be even more clean. There is a convention on how to name variables which store `boolean` values - they should always start with `is` prefix :

```
const isOrderValid = validateOrderName(orderName);
if (isOrderValid) { ... }
```

code/pizzabot.js
```diff
@@ -1,7 +1,95 @@
...
+  const orderQuantity = +prompt(`How many of ${orderName} do you want?`);
```
EmOlsz, line 84: Happy to see you remembered to convert values from `prompt` into number-type :)

code/style.css
```diff
@@ -0,0 +1,183 @@
+body {
+  background: #fff;
+  color: white;
+  margin: 0;
+  font-family: 'Lato', Sans-Serif;
```
EmOlsz, line 5: Nice that you remembered about font fallback!

code/style.css
```diff
@@ -0,0 +1,183 @@
...
+  text-align: center;
+}
+
+video {
+  display: block;
+  position: absolute;
+  z-index: -1;
+  margin: 0;
+  right: 0;
+  top: 0;
```
EmOlsz, line 41: Wrong order - semantically, it should go respectively `top`, `right`, `bottom`, `left`

code/style.css
```diff
@@ -0,0 +1,183 @@
...
+  align-items: center;
+  width: 100%;
+  margin: 50px 0;
+}
+
+/* form */
+
+form {
+display: flex;
+flex-direction: column;
```
EmOlsz, line 62: wrong indentation

code/style.css
```diff
@@ -0,0 +1,183 @@
...
+  width: 60%;
+ }
+
+.order {
+  margin: 15px;
+}
+
+.order input {
+  height: 50px;
+  width: 300px;
```
EmOlsz, line 80: You can achieve same result by setting two different properties, `font-size` and `padding` - that would be far more clean solution - `width` property can stay, but `height` is not necessary and should be avoided in this case

code/style.css
```diff
@@ -0,0 +1,183 @@
...
+  background: rgba(255, 255, 255, 0.8);
+  transition: 0.3s;
+}
+
+input::placeholder {
+  color: #000;
+}
+
+/* input::placeholder {
+} */
```
EmOlsz, line 102: For future check your code before you push it and delete all commented occurrences

code/style.css
```diff
@@ -0,0 +1,183 @@
...
+  position: relative;
+}
+
+img {
+  object-fit: cover;
+  height: 300px;
+  width: 60%;
+}
+
+#veg {
```
EmOlsz, line 173: We shouldn't style by id's - let's change them to class

code/style.css
```diff
@@ -0,0 +1,183 @@
...
+  right: 0;
+  top: 0;
+  min-width: 100%;
+  height: 600px;
+  object-fit: cover;
+}
+
+/****** MAIN ******/
+
+main {
```
EmOlsz, line 49: that's a lot of styling by selector - in ideal situation our CSS file should consist only from class styling

pull_request_template.md
```diff
@@ -1,5 +1,7 @@
-Link to my deployed project:
+Link to my deployed project: X
 
-My reflections on how this week's project turned out:
+My reflections on how this week's project turned out: 
 
 Things I'd like to have clarified or explained in more detail:
+
+Radio buttons ..
```
EmOlsz, line 7: What seems to be the problem? :)

EmOlsz (changes requested):
Congratulations Linda & Jonna, you made a very nice project :) Most of all, I'm happy to see clean, consequent JS logic you implemented - that's the most important part of Week 4. What's also great is that I can see your knowledge of HTML and CSS is pretty solid already - you've added semantic tags and consistent CSS styles, great job! Take a look on my comments, I hope they will help you in future projects.

Before I'll be able to accept you project, you need to add validation for orderQuantity as suggested in one comment.

Nevertheless, good job :)

lindahz:
> Congratulations Linda & Jonna, you made a very nice project :) Most of all, I'm happy to see clean, consequent JS logic you implemented - that's the most important part of Week 4. What's also great is that I can see your knowledge of HTML and CSS is pretty solid already - you've added semantic tags and consistent CSS styles, great job! Take a look on my comments, I hope they will help you in future projects.
> 
> Before I'll be able to accept you project, you need to add validation for orderQuantity as suggested in one comment.
> 
> Nevertheless, good job :)

Hello Maks! Thank you for clear and careful comments. Very much appreciated.  I have added validation to the number input and pushed the project back to github. I hope that you can approve now, or do I need to do another pull?

EmOlsz:
Hi Linda! Happy to hear that :) no need for another pull request, project is approved!

EmOlsz (approved):

PR #39, opened 2020-09-11
Pizza Bot - Hanna Skeppe and Therese Lindén

Hanna-Skeppe:
Link to my deployed project:
https://pizza-bot-hanna-and-therese.netlify.app/

We are proud of that we accomplished this project together without any prior knowledge of Javascript. 

We think that it would have helped us if we had gotten some more practice in basic Javascript before doing this project. Going into this project without a basic understanding of Javascript was quite challenging. Maybe the material and the lecture could have been more connected to the task.


code/index.html
```diff
@@ -1,12 +1,59 @@
...
-    <h1>This is your Pizza Bot</h1>
-  </body>
-</html>
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="stylesheet" href="./style.css">
+    <link href="https://fonts.googleapis.com/css2?family=Nanum+Gothic&display=swap" rel="stylesheet">
+    <link rel="icon" type="image/x-icon" href="favicon.ico">
+    <link rel="apple-touch-icon-precomposed" href="pizza-152-296065.png">
```
EmOlsz: Very good idea :)

code/index.html
```diff
@@ -1,12 +1,59 @@
...
-    <title>Project 4: Pizzabot</title>
-  </head>
-  <body>
-    <h1>This is your Pizza Bot</h1>
-  </body>
-</html>
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="stylesheet" href="./style.css">
```
EmOlsz: Thank you for `./` in file path!

code/index.html
```diff
@@ -1,12 +1,59 @@
...
+      <h1>Hey, welcome to your Pizza Bot!</h1>
+      <h2>Order your pizza here.</h2>
+      <!----Your name---->
+      <div class="input-element">
+          <label for="firstName">First name:</label>
+          <input
+              type="text"
+              id="firstName"
+              name="firstName"
+          />
```
EmOlsz: Great line breaks :)

code/index.html
```diff
@@ -1,12 +1,59 @@
...
+      </div>
+      <!---- Pizza selection---->
+      <div class="input-element">
+          <label for="pizzaSelection">Select pizza:</label>
+          <select id="pizzaSelection" name="pizzaSelection">
+              <option value="Vegetarian Pizza">Vegetarian Pizza</option>
```
EmOlsz: `value` attribute should be a string without whitespaces, written in `camelCase` - it's because this is not information for human, only for computer :

`<option value="thisIsInfoForBackEnd">This is part displayed in the browser for users</option>`

code/index.html
```diff
@@ -1,12 +1,59 @@
...
+              <option value="Hawaiian Pizza">Hawaiian Pizza</option>
+              <option value="Pepperoni Pizza">Pepperoni Pizza</option>
+              <option value="Kebab Pizza">Kebab Pizza</option>
+          </select>
+      </div>
+      <!---- Number of Pizzas ----->
+      <div class="input-element-slider">
+          <label class="slider-text" for="pizza-range">How many pizzas would you like?</label>
+          <input type="range" id="pizzaRangeSlider" class="pizza-range-slider" name="pizzaRangeSlider" min="0" max="8" value="1"
```
EmOlsz: Clever idea, I'm happy you did something different here than usual `<input type="number" />`

code/index.html
```diff
@@ -1,12 +1,59 @@
...
+      <!----Place order WILL ATT ORDER VISAS ONCLICK BUTTON---->
+      <div class="input-element">
+          <button type="button" onclick="return createPizzaOrder()">Place your order</button>
+      </div>
+    </form>
+      <!----Confirmation ---->
+    <div class="orderConfirmation">
+      <p id="orderConfirmationText"></p>
+        <!----PizzaSelection default image  ----->
+      <img id="pizzaPhoto" src="eatingPizza.jpg" alt ="picture of selected pizza">
```
EmOlsz: `./` missing

code/pizzabot.js
```diff
@@ -1,7 +1,109 @@
...
 
-//Put your Javscript code here:
+//console.log(vegetarian);
+//console.log(hawaiian);
+//console.log(pepperoni);
+
+/*Functions below */
+
+/*Check valid pizza name*/
+const validateOrderName = (orderName) =>{
```

`const validateOrderName = order => { ... }`

than

`const validateOrderName = (order) => { ... }`

code/pizzabot.js
```diff
@@ -1,7 +1,109 @@
...
+    if (orderName === vegetarian || orderName === hawaiian || orderName === pepperoni || orderName === kebab)
+    {
+        return true;
+    }
+    else return false;
```
Apart from returning given thing from function, `return` keyword does one more thing - it automatically finishes function. That's why we mustn't write anything below `return`, because that code will never be executed (of course in your case we have code below `return true`, but it's inside if statement, so it wouldn't enter second statement in the first place).
```
const validateOrderName = order => {
  if (orderName === vegetarian || orderName === pepperoni || orderName === hawaiian) {
    return true;
  }
  return false;
}
```
As you can see, we are not writing else statement anymore. We can do that thanks to behaviour mentioned above; if result of if statement is `true`, we will execute `return true` code from line 18 and finish function. If, on the other hand, result of it is `false`, we will never enter line 18 and jump straight into line 20. After executing line 20 we will finish function same way :)

code/pizzabot.js
```diff
@@ -1,7 +1,109 @@
...
+}
+//console.log(validateOrderName);
```
EmOlsz: Before pushing your code, check all commented stuff and delete it

code/pizzabot.js
```diff
@@ -1,7 +1,109 @@
...
+
+/*Calculate order total price*/
+const calculateTotalCost = (orderQuantity, pizzaPrice) =>{
+   const orderTotal = orderQuantity * pizzaPrice;
+   return orderTotal;
```
EmOlsz, line 22: Very nice :) Another tip from my side - since we have only one line inside `calculateTotalCost` function, we can write it in "arrow function shorthand" :

code/style.css
```diff
@@ -0,0 +1,176 @@
+
+* {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0; 
+}
+
+body {
+  font-family: 'Nanum Gothic', sans-serif;
+  background: url(./pizza-dough.jpg) no-repeat center center fixed;
```

background: url("./pizza-dough.jpg") no-repeat center center fixed;

code/style.css
```diff
@@ -0,0 +1,176 @@
...
+  align-items: center;
+  color: #f0e0e0;
+}
+h1 {
+  font-family: Lato;
+  font-weight: 100;
+  font-size: 48px;
+  -webkit-animation: neon1 1.5s ease-in-out infinite alternate;
+  -moz-animation: neon1 1.5s ease-in-out infinite alternate;
+  animation: neon1 1.5s ease-in-out infinite alternate;
```
EmOlsz, line 26: Great! :)

EmOlsz (changes requested):
Congratulations Hanna & Therese, good job this week! Your code in general clean and predictable, as it should be :) That's the most important thing to learn from last week - how to use functions and connect them to HTML, and you've done it well! I like your visualisation in HTML and range slider, very ambitious of you. Take a look on my comments, I hope they will help you to be even better programmers :)

Before I will approve your project, I need you to remove `console.log()` occurrences from code, please fix that as soon as possible.

Congratulations! :)

code/index.html
Hanna-Skeppe: Fixed it now :)

code/pizzabot.js
Hanna-Skeppe: Fixed it now :)

code/index.html
Hanna-Skeppe: Fixed now :)

code/pizzabot.js
Hanna-Skeppe: Ok, thanks for the info Maks, we fixed it :)

code/pizzabot.js
Hanna-Skeppe: Thanks for the tip Maks. We tried to change the function according to what you suggested, but didn't get it to work. We will look in to this in future projects.

Hanna-Skeppe:
We have made the changes that you suggested. Get back to us if it is something we missed :)

EmOlsz (approved):

PR #40, opened 2020-09-11
Sanna & Olof Pizza Bot Assignment

luuf81:
**Link to my deployed project:** https://sanna-olof-simple-pizzabot.netlify.app/

was fun to work together as a team. Was a bit unclear in the beginning how to do things according to the assignment. But things got clearer along the way. 

A bit unclear how to use the validateOrder function with if/else and get it to work with prompt in web browser. We submitted a question around this on StackOverflow: https://stackoverflow.com/c/technigo/questions/1447

We solved the problem by using the form instead.

Assignment was done by Sanna Dolck Wall & Olof Lindahl!

### DIFF
--- a/code/index.html
+++ b/code/index.html
@@ -6,7 +6,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
   <meta http-equiv="X-UA-Compatible" content="ie=edge" />
   <title>Project 4: Pizzabot</title>
-  <link rel="stylesheet" href="style.css">
+  <link rel="stylesheet" href="./style.css">
 </head>
 
 <body>
@@ -18,11 +18,17 @@
     <!-- Form for pizza order input -->
 
     <form class="input-form" onsubmit="handleSubmit(event)">
-      <label for="orderName">Which Pizza would you like?</label><br>
-      <input id="orderName" name="orderName" type="text" /><br>
-      <label for="orderQuantity">How many would you like?</label><br>
-      <input id="orderQuantity" name="orderQuantity" type="number" /><br>
-      <input id="submit" name="submit" type="submit" value="ORDER" />
+      <div>
+        <label for="orderName">Which Pizza would you like?</label>
+        <input id="orderName" name="orderName" type="text" />
+      </div>
+      <div>
+        <label for="orderQuantity">How many would you like?</label>
+        <input id="orderQuantity" name="orderQuantity" type="number" min="1"/>
+      </div>
+      <div>
+        <input id="submit" name="submit" type="submit" value="ORDER" />
+      </div>
     </form>
 
     <!-- text bubble showing order summary or incorrect pizza order -->
--- a/code/index.html
+++ b/code/index.html
@@ -8,10 +8,21 @@
     <link rel="stylesheet" href="style.css">
   </head>
   <body>
+    <div class="content">
     <h1>This is your Pizza Order Summary</h1>
 
+    <p id="welcome"></p>
+
+    <form class="input-form" onsubmit="handleSubmit(event)">
+      <label for="orderName">Which Pizza would you like?</label><br>
+      <input id="orderName" name="orderName" type="text"/><br>
+      <label for="orderQuantity">How many would you like?</label><br>
+      <input id="orderQuantity" name="orderQuantity" type="number"/><br>
+      <input id="submit" name="submit" type="submit" value="Submit"/> 
+    </form>
+
     <p id="orderSummary"></p>
-    
+  </div>
 
 
     <script src="pizzabot.js"></script>
--- a/code/index.html
+++ b/code/index.html
@@ -1,29 +1,36 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta charset="UTF-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
-    <title>Project 4: Pizzabot</title>
-    <link rel="stylesheet" href="style.css">
-  </head>
-  <body>
-    <div class="content">
-    <h1>This is your Pizza Order Summary</h1>
+
+<head>
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+  <title>Project 4: Pizzabot</title>
+  <link rel="stylesheet" href="style.css">
+</head>
+
+<body>
+  <div class="content">
+    <h1>Order your pizza here</h1>
 
     <p id="welcome"></p>
 
+    <!-- Form for pizza order input -->
+
     <form class="input-form" onsubmit="handleSubmit(event)">
       <label for="orderName">Which Pizza would you like?</label><br>
-      <input id="orderName" name="orderName" type="text"/><br>
+      <input id="orderName" name="orderName" type="text" /><br>
       <label for="orderQuantity">How many would you like?</label><br>
-      <input id="orderQuantity" name="orderQuantity" type="number"/><br>
-      <input id="submit" name="submit" type="submit" value="Submit"/> 
+      <input id="orderQuantity" name="orderQuantity" type="number" /><br>
+      <input id="submit" name="submit" type="submit" value="ORDER" />
     </form>
+
+    <!-- text bubble showing order summary or incorrect pizza order -->
 
     <p class="hidden-summary" id="orderSummary"></p>
   </div>
 
-    <script src="pizzabot.js"></script>
-  </body>
+  <script src="pizzabot.js"></script>
+</body>
+
 </html>
--- a/code/index.html
+++ b/code/index.html
@@ -21,9 +21,8 @@
       <input id="submit" name="submit" type="submit" value="Submit"/> 
     </form>
 
-    <p id="orderSummary"></p>
+    <p class="hidden-summary" id="orderSummary"></p>
   </div>
-
 
     <script src="pizzabot.js"></script>
   </body>
--- a/code/index.html
+++ b/code/index.html
@@ -5,8 +5,15 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <title>Project 4: Pizzabot</title>
+    <link rel="stylesheet" href="style.css">
   </head>
   <body>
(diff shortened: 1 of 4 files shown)


code/index.html
```diff
@@ -1,12 +1,36 @@
...
-    <h1>This is your Pizza Bot</h1>
-  </body>
-</html>
+
+<head>
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+  <title>Project 4: Pizzabot</title>
+  <link rel="stylesheet" href="style.css">
```

code/index.html
```diff
@@ -1,12 +1,36 @@
...
+<body>
+  <div class="content">
+    <h1>Order your pizza here</h1>
+
+    <p id="welcome"></p>
+
+    <!-- Form for pizza order input -->
+
+    <form class="input-form" onsubmit="handleSubmit(event)">
+      <label for="orderName">Which Pizza would you like?</label><br>
```
EmOlsz: We shouldn't use `br` tags - you can achieve same result (line break between `label` and `input`) by using `display: flex;` on `form` element or by wrapping each pair label-input into `div` and setting `flex` over there

code/pizzabot.js
```diff
@@ -1,7 +1,71 @@
...
+  "welcome"
+).innerHTML = `Hey! Happy to serve your pizza. On our menu we have ${vegetarian}, ${hawaiian} and ${pepperoni}.`;
+
+let orderName;
+let orderQuantity;
+
+// function to validate order. takes input from form and validates that pizza exist on the menu.
+// Returns true or false
+
+const validateOrderName = (validOrder) => {
```
EmOlsz, line 17: There is an arrow function convention which says that if we pass to our function only one argument, we can omit parentheses `()`, and rather write :

`const validateOrderName = validOrder => { ... }`

than

`const validateOrderName = (validOrder) => { ... }`

Important thing to notice - this convention should be applied ONLY when we pass one argument. If we would do this with multiple arguments, it would cause `SyntaxError` to occur.

Another thing I would change is name of parameter - `validOrder`. From semantic point of view, purpose of your function is to check whether order typed by user is valid, so argument passed into function could be valid, but don't have to. Naming it `validOrder` indicates that it's valid 100%, but we don't know that yet. As a sum up, `order` or `orderToBeValidated` would be better idea

code/pizzabot.js
```diff
@@ -1,7 +1,71 @@
...
+  return (
+    validOrder == vegetarian ||
+    validOrder == hawaiian ||
+    validOrder == pepperoni
```
EmOlsz, line 21: Congratulations, your idea to return boolean after this check is very clever!

code/pizzabot.js
```diff
@@ -1,7 +1,71 @@
...
+
+// Function to calculate cooking time based on order quantity
+
+const calculateCookingTime = (orderQuantity) => {
+  if (orderQuantity >= 1 && orderQuantity <= 3) {
+    return 10;
+  } else if (orderQuantity < 6) {
+    return 15;
+  } else {
+    return 20;
```
Apart from returning given thing from function, `return` keyword does one more thing - it automatically finishes function. That's why we mustn't write anything below `return`, because that code will never be executed (of course in your case we have code below `return 10` and `return 15`, but it's inside if statement, so it wouldn't enter second statement in the first place).
```
const calculateCookingTime = orderQuantity => {
  if (orderQuantity >= 1 && orderQuantity <= 3) {
    return 10;
  } else if (orderQuantity < 6) {
    return 15;
  }
  return 20;
}
```
As you can see, we are not writing else statement anymore. We can do that thanks to behaviour mentioned above; if result of if statement is `orderQuantity >= 1 && orderQuantity <= 3` or `orderQuantity < 6`, we will execute code from lines respectively 29 or 31 and finish function. If, on the other hand, result of it is higher than 6, we will never enter line 29 or 31 and jump straight into line 33. After executing line 33 we will finish function same way :)

code/pizzabot.js
```diff
@@ -1,7 +1,71 @@
...
+  }
+};
+
+// Function to calculate total cost based on order quantity
+
+const calculateTotalCost = (orderQuantity, pizzaPrice) => {
+  return orderQuantity * pizzaPrice;
+};
```
EmOlsz: Very nice :) Another tip from my side - since we have only one line inside calculateTotalCost function, we can write it in "arrow function shorthand" :

code/pizzabot.js
```diff
@@ -1,7 +1,71 @@
...
+
+// Function to toggle active css class of order summary bubble (to show/hide)
+
+const toggle = () => {
+  return this.classList.toggle("active-summary");
```
EmOlsz: There is couple of things wrong with this line of code :

1. You are using `return` keyword. You shouldn't do that, because you actually don't need to return anything from this function. What this function should do is just toggle class `active-summary`, but because you used `return` keyword, you are literally returning `undefined`. Why `undefined` though? Because it's result of calling build-in JS function `.toggle()`.
2. You are using `this` keyword. I encourage you to do `console.log()` and check what actually is `this`. You will see it's `Window` object, so not really the thing you thought it is. `this` keyword is far more complex subject, for sure not for a comment in code review. I encourage you to check that subject, however you should be aware it's not used so often anymore since ES6 introduced us to more clever ways of doing same stuff as with usage of `this`

code/pizzabot.js
```diff
@@ -1,7 +1,71 @@
...
+
+//Handles submit of order form. Will call each above function and generate either error message or order summary.
+
+const handleSubmit = (event) => {
+  event.preventDefault();
+  let orderName = document.getElementById("orderName").value;
+  let orderQuantity = document.getElementById("orderQuantity").value;
+
+  if (validateOrderName(orderName)) {
+    document.getElementById("orderSummary").classList = toggle;
```
EmOlsz, line 52: Check out in devtools what you actually added as class name to this HTML element

code/pizzabot.js
```diff
@@ -1,7 +1,71 @@
...
+const toggle = () => {
+  return this.classList.toggle("active-summary");
+};
+
+//Handles submit of order form. Will call each above function and generate either error message or order summary.
+
+const handleSubmit = (event) => {
+  event.preventDefault();
+  let orderName = document.getElementById("orderName").value;
+  let orderQuantity = document.getElementById("orderQuantity").value;
```
EmOlsz: I don't see any place where you are overwriting value of those variables, so in that case they should be defined with `const`

code/pizzabot.js
```diff
@@ -1,7 +1,71 @@
...
+    ).innerHTML = ` Great, I´ll get started on your ${orderName} right away, it will cost ${calculateTotalCost(
+      orderQuantity,
+      pizzaPrice
+    )} kr. The pizza(s) will take ${calculateCookingTime(
+      orderQuantity
+    )} minutes.`;
+  } else {
+    document.getElementById("orderSummary").classList = toggle;
+    document.getElementById("orderSummary").innerHTML =
+      "Please chooose a Pizza from the menu";
```
EmOlsz, line 64: Is it really necessary to do a line break over here? It's not so long line of code

code/style.css
```diff
@@ -0,0 +1,74 @@
+body {
+  background-color: #e9cec3;
+  font-family: Helvetica, Arial, Lucida, sans-serif;
```
EmOlsz, line 3: Happy to see font fallback :)

code/style.css
```diff
@@ -0,0 +1,74 @@
...
+  padding: 20px;
+  width: 25%;
+  border-radius: 25px;
+  text-align: center;
+  box-shadow: 10px 5px 5px #93786f;
+}
+
+.input-form input {
+  padding: 10px;
+  margin: 10px 0 10px 0;
```
EmOlsz, line 40: We can use shorthand here, `margin: 10px 0`

code/style.css
```diff
@@ -0,0 +1,74 @@
...
+  cursor: pointer;
```
EmOlsz, line 41: It's not really the best idea from accessibility point of view to add `cursor: pointer` to text inputs. They have good cursor set by default.
What you need to change is cursor for `<input type="submit" />` only

code/style.css
```diff
@@ -0,0 +1,74 @@
...
+}
+
+/* submit button */
+
+#submit {
```
EmOlsz, line 46: We should avoid styling by id as much as we can - only styling by classes is correct from semantic point of view

code/style.css
```diff
@@ -0,0 +1,74 @@
...
+  background-color: #dc9b81;
+  border-radius: 25px;
+  padding: 10px 20px 10px 20px;
```
EmOlsz, line 49: `padding: 10px 20px`

EmOlsz (changes requested):
Well done project Sanna & Olof ! I like your JS flow in general, apart from small mistakes you did a great job :) You HTML visualisation is nice as well, same as CSS implementation. Take a look on my comments, I hope they will help you to boost up your already proper knowledge and be better developers!

Before I'll approve your project, I need you to fix some things for me. Those are :
1. `br` tags should be replaced with proper CSS styles
2. Change logic behind your `toggle` function - for further information check out my comment
3. `let` -> `const`
4. Your code allowed me to type amount of pizzas as `-2` and it broke application - please fix that as well

code/index.html
luuf81: Fixed with just putting them in divs.

code/pizzabot.js
luuf81: Thanks. Will start using this :)

code/pizzabot.js
luuf81: Noted. Reason for using this was I wasn't sure how the function would know what to change (since we´re not passing any arguments to the function). It works now when omitting both `return` and `this` but still a bit puzzled how it knows its making the assignment to the right element?

code/pizzabot.js
luuf81: Noted. The logic I was that the orderName could change but really it doesn´t with the functionality we implemented.

luuf81:
Also added validation to the form so what only numbers >= 1 are allowed. So now all fixes that was asked for are made.

code/pizzabot.js
EmOlsz: I see. Happy to hear that you changed it at least :) Still, I don't want you to stay with doubts, so here is how I would implement this function :

This is function declaration
`const toggle = element => element.classList.toggle("active-summary");`

This is function call
`toggle(document.getElementById("orderSummary"))`

What do you think about that?

EmOlsz (approved):

EmOlsz:
Approved, great job :)

PR #41, opened 2020-09-11
Project 3-Pizzabot Jessica Panditha & Henrike Wiemker

jespa374:
Link to my deployed project:
https://mammamiapizza.netlify.app/

This project went well eventhough it was sometimes difficult with functions. Overall we are happy with the result. 

Not at the moment.


code/index.html
```diff
@@ -4,9 +4,30 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
-    <title>Project 4: Pizzabot</title>
+    <link rel="stylesheet" href="./style.css">
```
EmOlsz, line 7: Happy to see `./` in path to file!

code/index.html
```diff
@@ -4,9 +4,30 @@
...
+    <section class="main">
+    <div><p id="welcome"></p></div>
+    <form class="pizzabotform" onsubmit="handleSubmit(event)">
+      <label for="pizzaName">Enter the name of the pizza you want to order today.</label>
+      <input id="pizzaName" name="pizzaName" type="text">
+      <!--<input id="submitPizza" name="submitPizza" type="submit">-->
+      <label for="pizzaQuantity">How many pizzas do you want?</label>
+      <input id="pizzaQuantity" name="pizzaQuantity" type="number" min="0" max="20">
+    </form>
+    <button onclick="handleSubmit(event)">Order Now</button>
```
EmOlsz, line 23: No need to add `onclick` event since you already added `onsubmit` to your `form`. It's an alternative - either you choose to add `onclick` event to `button` and there is no need for `form` element at all, or you choose `onsubmit` event in `form` tag to handle everything and then your button just have to have `type="submit"` attribute

code/index.html
```diff
@@ -4,9 +4,30 @@
...
+    <div><p id="orderConfirmation"></p></div>
+    <div class="pizzacontainer">
+      <img src="Images/hawaiianpizza.jpg">
```
EmOlsz: all folders should be written with lowercase letters only. Proper path would be :

``

Also, let's not forget about `alt` attribute, `./` in path and proper closing of our tag, so the final result should look like this :

`Hawaiian pizza`

code/index.html
```diff
@@ -4,9 +4,30 @@
...
+      <img src="Images/pepperonipizza.jpg">
+      <img src="Images/vegetarianpizza.jpg">
+    </div>
+  </section>
+    <script src="pizzabot.js"></script>
```
EmOlsz: Missing `./` in file path

code/pizzabot.js
```diff
@@ -2,6 +2,54 @@ const vegetarian = "Vegetarian Pizza"
 const hawaiian = "Hawaiian Pizza"
 const pepperoni = "Pepperoni Pizza"
 
-const pizzaPrice = 80
+const pizzaPrice = 80;
 
 //Put your Javscript code here:
+console.log(vegetarian,hawaiian,pepperoni);
```
EmOlsz: Before submitting your project, always remember to delete all occurrences of `console.log()`

code/pizzabot.js
```diff
@@ -2,6 +2,54 @@ const vegetarian = "Vegetarian Pizza"
...
+
+
+document.getElementById("welcome").innerHTML = `Hey! Happy to serve your pizza. On our menu we have ${vegetarian}, ${hawaiian} and ${pepperoni}`
```
EmOlsz: You seem to add semicolons `;` sometimes and sometimes not - let's be consequent, you either need to add them everywhere or not use them at all

code/pizzabot.js
```diff
@@ -2,6 +2,54 @@ const vegetarian = "Vegetarian Pizza"
...
+
+const calculateCookingTime = (orderQuantity) => {
```

`const calculateCookingTime = orderQuantity => { ... }`

than

`const calculateCookingTime = (orderQuantity) => { ... }`

code/pizzabot.js
```diff
@@ -2,6 +2,54 @@ const vegetarian = "Vegetarian Pizza"
...
+    if (orderQuantity <= 2 ) {
+       return 10;
+    } else if (orderQuantity >= 3 && orderQuantity <= 5) {
+       return 15;
+    } else {
+       return 20;
```
EmOlsz, line 16: I was able to enter value lower than 0 and it led to bizarre situation when I had to pay `-169kr` for my order - let's fix that :)

code/pizzabot.js
```diff
@@ -2,6 +2,54 @@ const vegetarian = "Vegetarian Pizza"
...
+    }
+}
+
+
+const validateOrderName = (orderName) => {
+    if (orderName === vegetarian || orderName === hawaiian || orderName === pepperoni) {
+        return true 
+    } else { 
+        return false
+    }
```
Apart from returning given thing from function, `return` keyword does one more thing - it automatically finishes function. That's why we mustn't write anything below return, because that code will never be executed (of course in your case we have code below return true, but it's inside if statement, so it wouldn't enter second statement in the first place).
```
const validateOrderName = orderName => {
  if (orderName === "vegetarian" || orderName === "hawaiian" || orderName === "pepperoni") {
    return true;
  }
  return false;
}
```
As you can see, we are not writing else statement anymore. We can do that thanks to behaviour mentioned above; if result of if statement is `true`, we will execute `return true` code from line 26 and finish function. If, on the other hand, result of it is `false`, we will never enter line 26 and jump straight into line 28. After executing line 28 we will finish function same way :)

code/pizzabot.js
```diff
@@ -2,6 +2,54 @@ const vegetarian = "Vegetarian Pizza"
...
+};
+
+const calculateTotalCost = (orderQuantity, pizzaPrice) => {
+    return orderQuantity*pizzaPrice;
+}
```
EmOlsz, line 33: Very nice :) Another tip from my side - since we have only one line inside calculateTotalCost function, we can write it in "arrow function shorthand" :

`const calculateTotalCost = (orderQuantity, pizzaPrice) => orderQuantity * pizzaPrice;`

code/pizzabot.js
```diff
@@ -2,6 +2,54 @@ const vegetarian = "Vegetarian Pizza"
...
+
+const handleSubmit = (event) => {
+    event.preventDefault();
+    const orderName = document.getElementById("pizzaName").value
+        /*prompt('Enter the name of the pizza you want to order today.');*/
+    
+    validateOrderName(orderName); 
+    if (validateOrderName(orderName) === true) {
```
EmOlsz: There is no need to call your function in line 41 if you do it in line 42 as well. Let me explain to you what is happening over here :

```
41.        validateOrderName(orderName); 
42.        if (validateOrderName(orderName) === true) { ... }
```

Line 41 calls `validateOrderName()` function. That function only purpose is to return `true` or `false`. But where do you return it? Nowhere. Your function is called, returns boolean value and... nothing. `return` keyword main purspose is to assing it to variable or to compare it with something else. Let's assume that our function will return `true` - your line of code literally looks like this for JavaScript: 

```
41.        true;
42.        if (validateOrderName(orderName) === true) { ... }
```

Only in line 42. you are actually using `validateOrderName()` function in any sort of way - you are comparing value returned from it with `true`. Let's once again assume our function will return `true`. This is how JavaScript sees this code :

```
41.        true;
42.        if (true === true) { ... }
```

Let's fix your code a bit then. First of all, we don't need that `true` code in line 41, it doesn't do anything at all. To check that try to create new JS project with `.js` file and type `true` over there. It's literally same as you would not call it at all - `boolean` values should be used to compare stuff mostly.
The cleanest approach over here would be creating variable which only purpose will be storing value returned from `validateOrderName()` function. Later on, we will use this variable in if statement. Convention for variables storing boolean values is that they should always begin with prefix `is`. It can look like this :

```
41.        const isOrderNameValid = validateOrderName(orderName);
42.        if (isOrderNameValid === true) { ... }
```

However, we can simplify our code even more. Since `isOrderNameValid` is already `boolean` value, we don't need to compare it with `true` - result of that comparison would be just another `boolean`, but we already have it! So, we can simplify it like this :

```
41.        const isOrderNameValid = validateOrderName(orderName);
42.        if (isOrderNameValid) { ... }
```

code/pizzabot.js
```diff
@@ -2,6 +2,54 @@ const vegetarian = "Vegetarian Pizza"
...
+        const orderQuantity = document.getElementById("pizzaQuantity").value;
+        /*prompt(`How many of ${orderName} do you want?`);*/
```
EmOlsz: commented code

code/style.css
```diff
@@ -0,0 +1,42 @@
...
+header {
+    display: flex;
+    justify-content: center;
+}
+
+.main {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    /*justify-content: center;*/
```
EmOlsz: commented code

EmOlsz (changes requested):
Great job Jessica and Henrike! You've prepared fine project this week :) Your visualisation shows that you know how to handle proper, semantic HTML and CSS structure. Your JS flow is logical and clean (apart from small changes I mentioned in comments). Congratulations :)

Before I will be able to approve your application, I need to see some changes in code. Those are :
1. Decide which event in regards to `<form />` you want to keep - it's either going to be `onsubmit` or `onclick`. Both are too much.
2. Delete `console.log()`and commented code occurrences
3. Decide whether you want to use semicolons or not
4. Validate your numeric input so user cannot pass values lower than 0

EmOlsz (approved):

PR #42, opened 2020-09-11
Katarina's and Annika's final project!

Katsiez:
Link to my deployed project: https://github.com/Katsiez/project-pizza-bot
Netlify link to be found inside README

Working in pairs has been amazing! We were both kind of scared of how it might turn out, but everything went smoothly and we realized how much we have completed in such a short time, thanks to the fact we were two!

I would like to know how to include images after the order has been placed.
And on that same page, I would like to have a button that when clicked takes the user back to the order page.
Also, I'd like to know how to clear the input. I tried doing that a few times but it never worked.


code/index.html
```diff
@@ -5,8 +5,69 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <title>Project 4: Pizzabot</title>
+    <link rel="stylesheet" href="style.css" />
```

code/index.html
```diff
@@ -5,8 +5,69 @@
...
+    <link rel="icon" type="image/png" href="./img/pizza.png" />
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/2.0.2/jquery.min.js"></script>
   </head>
-  <body>
-    <h1>This is your Pizza Bot</h1>
+ 
+<body id="orderPlaced">
```
EmOlsz: wrong indentation when comparing to `head`

code/index.html
```diff
@@ -5,8 +5,69 @@
...
+
+    <!-- H1 NEON TITLE   -->
+    <div class="neon">
+      <h1 class="green">This is your Pizza Bot</h1>
+    </div>
+    
+    <!-- CHATBOT FORM -->
+    <section class="chatbot-form">
```
EmOlsz, line 22: This whole section has a lot of wrong indents, moreover I saw couple places where you open `div`, but there is no closing tag - that needs to be fixed

code/index.html
```diff
@@ -5,8 +5,69 @@
...
+      <div id="chatbot">
+        <div class="chat">
+          <h2 class="h2-chat">Hey, happy to serve your pizza!</h2>
+          <p>On our menu we have: Vegetarian Pizza, Hawaiian Pizza, Pepperoni Pizza and Margherita Pizza.<p>
+            
+            <div class="form-group">
+              <label for="firstName"><h3 class="h3-order">To start, enter your name</h3></label>
+              <input
+                  id="firstName"
+                  type="text"
```
EmOlsz, line 33: Happy to see this line breaks :)

code/pizzabot.js
```diff
@@ -1,7 +1,47 @@
...
+        orderTime = 10
+    } else if (orderQuantity >= 3 && orderQuantity <= 5) {
+        orderTime = 15
+    } else if (orderQuantity >= 6) {
+        orderTime = 20
+    }
+}
+
+let validateOrderName = (orderName, orderQuantity) => {
+    event.preventDefault();
```
EmOlsz: So are we using semicolons in this project or not?
It's up to you whether you use them or not, what's most important is consistence - you either use it everywhere, or not at all

code/pizzabot.js
```diff
@@ -1,7 +1,47 @@
...
 const vegetarian = "Vegetarian Pizza"
 const hawaiian = "Hawaiian Pizza"
 const pepperoni = "Pepperoni Pizza"
+const margherita = "Margherita Pizza"
 
 const pizzaPrice = 80
 
-//Put your Javscript code here:
+
+const calculateTotalCost = (orderQuantity) => {
```

`const calculateTotalCost = orderQuantity => { ... }`

than

`const calculateTotalCost = (orderQuantity) => { ... }`

code/pizzabot.js
```diff
@@ -1,7 +1,47 @@
...
+    orderTotal = (orderQuantity * pizzaPrice)
```
EmOlsz: No need for extra pair of parentheses here, `orderTotal = orderQuantity * pizzaPrice;` is OK :)

code/pizzabot.js
```diff
@@ -1,7 +1,47 @@
...
+    } else if (orderQuantity >= 3 && orderQuantity <= 5) {
+        orderTime = 15
+    } else if (orderQuantity >= 6) {
+        orderTime = 20
+    }
+}
+
+let validateOrderName = (orderName, orderQuantity) => {
+    event.preventDefault();
+    if (orderName === vegetarian || orderName === hawaiian || orderName === pepperoni) {
```
EmOlsz: At this moment if - as a user - I will type `vegetarian pizza` instead of `Vegetarian Pizza`, that value will not be validated. Maybe it's worth adding conditional like this ?

`if (orderName.toLowerCase() === vegetarian.toLowerCase() || ... )`

code/pizzabot.js
```diff
@@ -1,7 +1,47 @@
...
+        calculateTotalCost(orderQuantity)
+        calculateCookingTime(orderQuantity)
+        document.getElementById("orderPlaced").innerText = `Thanks ${firstName.value}! You have ordered ${orderQuantity} of the ${orderName}, your total is ${orderTotal} kr and your order will take ${orderTime} minutes.`
+    } else if (orderName === margherita) {
+        document.getElementById("final").innerText = `Oh no! ${margherita} is sadly out of stock. Please order something else!`
```
EmOlsz, line 31: Clever ;)

code/style.css
```diff
@@ -0,0 +1,218 @@
...
+
+body {
+  width: 100%;
+  height: 100%;
+  margin: 0;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  background-image: url(./img/bluepizza.jpg);
```

`background-image: url("./img/bluepizza.jpg");`

code/style.css
```diff
@@ -0,0 +1,218 @@
...
+  background-position: center;
+}
+
+#orderPlaced {
```
EmOlsz, line 22: We should avoid styling by ids - let's change those to classes

code/style.css
```diff
@@ -0,0 +1,218 @@
...
+    background-position: 0% 50%;
+  }
+  50% {
+    background-position: 100% 50%;
+  }
+  100% {
+    background-position: 0% 50%;
+  }
+}
+@-moz-keyframes AnimationName {
```
EmOlsz: Very nice! :)
Just one thing - animation name should be written lowercase

EmOlsz (changes requested):
Nice project Katarina & Annika, congratulations :) Not everything is ideal, but that doesn't change the fact that both your JS logic and HTML/CSS structure are done well! JS part is far more important this week and you made it work correctly! Take a look on my comments for further reading, hope that will help :)

Before I will accept you project officially, I need to see some changes in code. Those are :
1. Fix indentation throughout HTML and CSS
2. Decide whether you want or not to use semicolons

Katsiez:
Hi Maks,
Thank you for your comments! I tried to keep the indentation in line, even
downloaded an extension in VS code that keeps the indentation right upon
save. Any tips on how I could fix this further with keyboard shortcuts? I’m
on Mac.

Thank you again!
Katarina

EmOlsz (approved):

PR #43, opened 2020-09-11
Evelina & Agnes pizzabot 

aagneshansson:
Link to my deployed project:
https://getyourpizza.netlify.app/

We had a great collaboration and lots of fun. It was a really challenging task but we manage to solve it together.

How to connect functions in Javascript together with html and how to hide and show items(like a chatbot).


code/index.html
```diff
@@ -4,9 +4,79 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <link rel="stylesheet" href="style.css" />
```
EmOlsz, line 7: Let's not forget about `./` in path to file

code/index.html
```diff
@@ -4,9 +4,79 @@
...
+          <div class="question-container">
+            <p id="pizzaName"></p>
+
+            <!-- Input form for what pizza they would like to order -->
+            <input
+              type="text"
+              id="pizzaNameInput"
+              name="pizza"
+              placeholder="Name of pizzas goes here"
+            />
```
EmOlsz, line 33: Happy to see these line breaks :)

code/index.html
```diff
@@ -4,9 +4,79 @@
...
+            <p id="pizzaQuantity"></p>
+
+            <!-- Input form for ask how many pizzas-->
+            <input
+              type="number"
+              id="pizzaQuantityInput"
+              name="pizzanumber"
+              placeholder="Number of pizzas goes here"
+              min="1"
+              max="100"
```
EmOlsz, line 47: I'm glad you remembered about `min` attribute, so there is no option to order -2 pizzas!

code/index.html
```diff
@@ -4,9 +4,79 @@
...
+            />
+          </div>
+
+          <!-- Place the order-->
+          <div class="submit-container">
+            <button
+              class="form-submit"
+              type="submit"
+              id="submit"
+              name="submit"
```
EmOlsz, line 56: No need to add `name` attribute here, `<input type="submit" />`, or, in your case, `<button type="submit" />` is a special part of each form, which is only responsible for submitting form. When you give `name` attribute to it, you are somehow saying "I want to store value from that field and send it to back-end". But there is no value important for back-end here, right?

code/index.html
```diff
@@ -4,9 +4,79 @@
...
+
+        <!-- Message that appears after the order is placed -->
+        <div class="order-message-container">
+          <p id="finalOrderMessage"></p>
+        </div>
+      </section>
+    </main>
+    <footer>
+      <div class="footer-info-container">
+        <h2 class="footer-info-header">
```
EmOlsz, line 72: I really like names you have to your classes, they are very descriptive :)

code/index.html
```diff
@@ -4,9 +4,79 @@
...
+          Made by Agnes Harrysson Hansson & Evelina Berg
+        </h2>
+        <p class="footer-info-text">A Technigo bootcamp project</p>
+      </div>
+    </footer>
+
+    <script src="pizzabot.js"></script>
```
EmOlsz, line 79: `./` over here as well

code/pizzabot.js
```diff
@@ -1,7 +1,98 @@
...
+const hawaiian = "Hawaiian"
+const pepperoni = "Pepperoni"
 const pizzaPrice = 80
 
-//Put your Javscript code here:
+
+// 2. Functions
+
+// Function for validating ordername
+const validateOrderName = (orderName) => {
```
EmOlsz, line 12: There is an arrow function convention which says that if we pass to our function only one argument, we can omit parentheses `()`, and rather write :

`const validateOrderName = orderName => { ... }`

than

`const validateOrderName = (orderName) => { ... }`

code/pizzabot.js
```diff
@@ -1,7 +1,98 @@
...
+    if (orderName.toLowerCase() === vegetarian.toLowerCase()) {
+        return true;
+    } else if (orderName.toLowerCase() === hawaiian.toLowerCase()) {
```
EmOlsz, line 15: That's great that you remembered about validating text user types into fields and make them lowercase - thanks to that we can accept both lower and upper case pizza names.
Nevertheless, we could simplify this conditional by typing only :
```
if (orderName.toLowerCase() === vegetarian.toLowerCase() || orderName.toLowerCase() === hawaiian.toLowerCase() || orderName.toLowerCase() === pepperoni.toLowerCase()) {
        return true;
}

File: code/pizzabot.js [hunk 9]
```diff
@@ -1,7 +1,98 @@
...
+    } else {
+        alert('Select a pizza from the menu!');
+        return false;
+    }
+};
+
+// Function for calculating total cost
+const calculateTotalCost = (orderQuantity, pizzaPrice) => {
+    let orderTotal = orderQuantity * pizzaPrice;
+    return orderTotal;
```
EmOlsz, line 28: We can simplify this function. Since you are creating `let` variable (should be `const` btw because you are not overwriting its value anywhere) and returning it very next line, we could declare it like this 
```
const calculateTotalCost = (orderQuantity, pizzaPrice) => {
    return orderQuantity * pizzaPrice;
}
```
Another tip from my side - since we have only one line inside `calculateTotalCost` function, we can write it in "arrow function shorthand" :

`const calculateTotalCost = (orderQuantity, pizzaPrice) => orderQty * pizzaPrice;`

We can do that because in arrow functions we can omit curly brackets (`{}`) and `return` keyword if we only have one line of code inside. At the end, this snippet of code is exactly the same as the first one I wrote above.

code/pizzabot.js
```diff
@@ -1,7 +1,98 @@
...
+};
+
+// Function for calculating total cooking time
+const calculateCookingTime = (orderQuantity) => {
+    let orderTime = 0; 
+    if (orderQuantity <= 2) {
+        orderTime = 10;
+        return orderTime;
```
EmOlsz, line 36: Again, you don't have to create another variable if its only purpose is to be returner - you can write it like this :
```
if (orderQuantity <= 2) {
    return 10;
}
```

code/pizzabot.js
```diff
@@ -1,7 +1,98 @@
...
+};
+
+// Function for input value from pizza name
+const getOrderName = () => {
+    return document.getElementById('pizzaNameInput').value;
+};
+
+// Function for input value from pizza quantity
+const getOrderQuantity = () => {
+    return document.getElementById('pizzaQuantityInput').value;
```
EmOlsz, line 68: Very well usage of `return` keyword :)

code/pizzabot.js
```diff
@@ -1,7 +1,98 @@
...
+};
+
+// The chatbot / form
+
+// Welcome message 
+const firstMessage = document.querySelector('.welcome-message-container'); 
+const welcomeMessage = `Hey! Happy to serve your pizza!`;
+firstMessage.innerHTML = `<div>${welcomeMessage}</div>`
+console.log(welcomeMessage);
```
EmOlsz: Before pushing your code to github, always check if there are any `console.log()` and delete them - it's considered bad practise to push them. Everyone can see your code out here and it's pretty messy

code/pizzabot.js
```diff
@@ -1,7 +1,98 @@
...
+askPizzas();
+askNumber();
+
+// Handle user input
+const handleOrder = (event) => {
+  event.preventDefault();
+  const orderName = getOrderName();
+  const orderQuantity = getOrderQuantity();
+  const isValidOrder = validateOrderName(orderName);
+  if (isValidOrder === true) {
```
EmOlsz, line 88: Do you remember when I was explaining to all of you how does if conditional works? Let's say that `isValidOrder` is equal to `true`. At the end, JavaScript sees this code :
```
 if (isValidOrder === true) { ... }
```
like this :
```
 if (true === true) { ... }
```
`true` is equal `true`, so after JavaScript checks it, the same line of code looks like this :
```
if (true) { ... }
```
See? So you could achieve that writing you code simpler, like this :
```
if (isValidOrder) { ... }
```
It's actually very nice you named your variable `isValidOrder`, because it's another convention - basically, all variables which store `boolean` values should be named with prefix `is`. So, `isValidOrder`, `isButtonClicked`, `isFormSubmitted` etc. Thanks to that when we write conditional like the one above, it actually makes sense both for computer and for human :) It's like a question - is order valid? If yes, do something. If not, do something else.

code/pizzabot.js
```diff
@@ -1,7 +1,98 @@
...
+        orderTime = 10;
+        return orderTime;
+
+    } else if (orderQuantity < 6) {
+        orderTime = 15;
+        return orderTime;
+
+    } else {
+        orderTime = 20;
+        return orderTime;
```
EmOlsz, line 44: Great job :)
Apart from returning given thing from function, `return` keyword does one more thing - it automatically finishes function. That's why we mustn't write anything below return, because that code will never be executed (of course in your case we have code below `return true`, but it's inside if statement, so it wouldn't enter second statement in the first place).
```
const calculateCookingTime = (orderQuantity) => {
    let orderTime = 0; 
    if (orderQuantity <= 2) {
        orderTime = 10;
        return orderTime;

    } else if (orderQuantity < 6) {
        orderTime = 15;
        return orderTime;
    }
    orderTime = 20;
    return orderTime;
}
```
As you can see, we are not writing `else` statement anymore. We can do that thanks to behaviour mentioned above; if result of if statement is lower than 6, we will execute code from one of two first conditionals and finish function. If, on the other hand, result of it is equal or higher than 6, we will never enter any of the `if` conditionals and execute only two last lines of this functon (so, `orderTime = 20` and `return orderTime`). After executing them, we will finish function same way :)

code/pizzabot.js
```diff
@@ -1,7 +1,98 @@
...
+    }
+};
+
+//Function for asking what kind of pizza user would like to order 
+const askPizzas = () => {
+    const orderName = `On our menu we have ${vegetarian}, ${hawaiian} and ${pepperoni}. Choose which pizza you want.`
+    document.getElementById("pizzaName").innerHTML = orderName;
```
EmOlsz, line 51: You seem to use semicolons only when you want. We have to be consequent - either use them everywhere or don't use them at all

code/style.css
```diff
@@ -0,0 +1,118 @@
+* {
+  box-sizing: border-box;
+}
+
+@import url(https://fonts.googleapis.com/css?family=Noto+Sans);
```
EmOlsz, line 5: Imports should be placed above everything else, at the very top of CSS files

code/style.css
```diff
@@ -0,0 +1,118 @@
...
+  background-color: rgba(255, 255, 255, 0.8);
+  width: 50%;
+  font-family: "Playfair Display", serif;
+  margin: 50px;
+  padding: 20px;
+  font-size: 15px;
+}
+
+.welcome-message-container {
+  margin-top: 0px;
```
EmOlsz, line 42: If value is equal `0px`, we can omit `px` and just write `margin-top: 0;`

code/style.css
```diff
@@ -0,0 +1,118 @@
...
+  align-items: center;
+  padding: 20px;
+}
+
+.form-submit {
+  width: 150px;
+  padding: 15px;
+  font-size: 15px;
+  cursor: pointer;
+  border: 1px solid 0.5;
```
EmOlsz: `0.5` of what - `px`, `meters`, `pizzas`? ;)
Btw, most browsers doesn't detect values lower than `1px` since it's the smallest unit, so we should use only fixed numbers here

EmOlsz (changes requested):
Congratulations Evelina & Agnes, nice job :) Your code is consequent and clean, I enjoyed analysing it. I've added some comments below, but most of them are tips. Please, consider them as very important, because I will be paying a lot of attention to them in future weeks. Apart from that, happy to see this quality :)

Before I'll be able to officially approve your project, you need to add some changes. Those are :
1. Delete `type="submit"` from `<button />`
2. Decide whether you want to use semicolons or not and fix your code according to that
3. Delete `console.log()`
4. Change `border: 1px solid 0.5`

EmOlsz (approved):

PR #44, opened 2020-09-12
Sara and Liza's pizza bot.

ssundblom:
Link to my deployed project:

https://pedantic-feynman-2546dc.netlify.app/

(here is the git push history: https://github.com/ssundblom/pizzabot2)

We copied code from a separate repo on 12/9 - therefor the big commit.

I'm just happy that we git it to work. It has been a long and hard week and we got stuck in the middle of it and could't find our way out.

How to connect HTML and JS.

(Sara Sundblom & Liza Minican)


code/index.html
```diff
@@ -1,12 +1,29 @@
...
-    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
-    <title>Project 4: Pizzabot</title>
-  </head>
-  <body>
-    <h1>This is your Pizza Bot</h1>
-  </body>
-</html>
+
+<head>
+    <link rel="stylesheet" href="style.css">
```

code/index.html
```diff
@@ -1,12 +1,29 @@
...
+    <div class="container">
+        <h1>Liza's Pizza Place</h1>
+
+        <h2>On our menu we have Vegetarian Pizza, Hawaiian Pizza,
+            Pepperoni Pizza. </h2>
+
+        <h2 id="botSays"></h2>
+
+        <div id="container">
+            <input type="text" id="answer" value="" placeholder="enter pizza name">
```
EmOlsz, line 21: No need to write `value=""`, just omit this attribute next time

code/index.html
```diff
@@ -1,12 +1,29 @@
...
+            <input type="button" id="submitButton" value=">" onclick="submitAnswer()">
+        </div>
+    </div>
+
+    <script src="pizzabot.js"></script>
```
EmOlsz: `./` over here as well

code/pizzabot.js
```diff
@@ -1,7 +1,89 @@
...
-const hawaiian = "Hawaiian Pizza"
-const pepperoni = "Pepperoni Pizza"
+const vegetarian = "Vegetarian Pizza";
+const hawaiian = "Hawaiian Pizza";
+const pepperoni = "Pepperoni Pizza";
 
-const pizzaPrice = 80
+const pizzaPrice = 80;
+
+let botSays = document.getElementById('botSays');
```
EmOlsz: I see you are not overwriting value if this variable anywhere, let's use `const` instead.

code/pizzabot.js
```diff
@@ -1,7 +1,89 @@
...
+botSays.innerHTML = 'What would you like to order?';
+
+// make inputbox actice (the mouse/cursor is there in beginning)
+document.getElementById("answer").focus();
+
+// used to check if valid pizza is ordered
+let orderedPizza = "";
```
EmOlsz: This is good usecase for `let` :)

code/pizzabot.js
```diff
@@ -1,7 +1,89 @@
...
+const pepperoni = "Pepperoni Pizza";
 
-const pizzaPrice = 80
+const pizzaPrice = 80;
+
+let botSays = document.getElementById('botSays');
+botSays.innerHTML = 'What would you like to order?';
+
+// make inputbox actice (the mouse/cursor is there in beginning)
+document.getElementById("answer").focus();
```
EmOlsz: Nice, very user-friendly :)

code/pizzabot.js
```diff
@@ -1,7 +1,89 @@
...
+
+// used to check if valid pizza is ordered
+let orderedPizza = "";
+
+const isValidPizza = (pizza) => {
```
EmOlsz, line 16: There is an arrow function convention which says that if we pass to our function only one argument, we can omit parentheses `()`, and rather write :

`const isValidPizza = pizza => { ... }`

than

`const isValidPizza = (pizza) => { ... }`

Important thing to notice - this convention should be applied ONLY when we pass one argument. If we would do this with multiple arguments, it would cause `SyntaxError` to occur.

Another thing about naming conventions. Functions should be named by this pattern : `validatePizza`, `checkInput`, `showForm`, `submitInput` etc. Basically, a pair of verb and noun in this order.

`isValidPizza` is actually a convention, but for variables that store `boolean` values, not functions. Good example of when to use it would be 
```
const isPizzaValid = validatePizza('pepperoni');
```

code/pizzabot.js
```diff
@@ -1,7 +1,89 @@
...
+    if (vegetarian.toLowerCase() === pizza.toLowerCase() ||
```
EmOlsz, line 17: Great job with checking lowercase values :)

code/pizzabot.js
```diff
@@ -1,7 +1,89 @@
...
+        hawaiian.toLowerCase() === pizza.toLowerCase() ||
+        pepperoni.toLowerCase() === pizza.toLowerCase()) {
+        return true
+    } else {
+        return false
```
EmOlsz, line 22: Regarding clean code, there is also another convention, let me introduce you to it.
Apart from returning given thing from function, `return` keyword does one more thing - it automatically finishes function. That's why we mustn't write anything below return, because that code will never be executed (of course in your case we have code below return true, but it's inside if statement, so it wouldn't enter second statement in the first place).
```
const validatePizza = pizza => {
    if (vegetarian.toLowerCase() === pizza.toLowerCase() ||
        hawaiian.toLowerCase() === pizza.toLowerCase() ||
        pepperoni.toLowerCase() === pizza.toLowerCase()) {
        return true
    }
    return false
}
```
As you can see, we are not writing `else` statement anymore. We can do that thanks to behaviour mentioned above; if result of if conditional is `true`, we will execute code from line 18 onwards and finish function at line 20. If, on the other hand, result of conditional was `false`, we will never enter line 18 and jump straight into line 22. After executing line 22 we will finish function same way :)

code/pizzabot.js
```diff
@@ -1,7 +1,89 @@
...
+    }
+}
+
+const calculateTotalCost = (orderQuantity, pizzaPrice) => {
+    let totalCost = orderQuantity * pizzaPrice
+    return totalCost
```
EmOlsz, line 28: We can simplify this function. Since you are creating `let` variable (btw it should be `const`, because you are not overwriting its value anywhere) and returning it very next line, we could declare it like this
```
const calculateTotalCost = (orderQuantity, pizzaPrice) => {
    return orderQuantity * pizzaPrice;
}
```
Another tip from my side - since we have only one line inside `calculateTotalCost` function, we can write it in "arrow function shorthand" :

`const calculateTotalCost = (orderQuantity, pizzaPrice) => orderQty * pizzaPrice;`

We can do that because in arrow functions we can omit curly brackets (`{}`) and `return` keyword if we only have one line of code inside. At the end, this snippet of code is exactly the same as the first one I wrote above.

code/pizzabot.js
```diff
@@ -1,7 +1,89 @@
...
+    let answer = document.getElementById("answer").value;
+
+    // If no pizza, or no valid pizza, has been ordered..
+    if (orderedPizza === "") {
+
+        // .. check if the pizza they want is valid
+        let validPizza = isValidPizza(answer)
+
+        if (validPizza) {
+            orderedPizza = answer;
```
EmOlsz: You seem to be using semicolons only when you want. We have to be consequent - either use them everywhere or don't use them at all

code/pizzabot.js
```diff
@@ -1,7 +1,89 @@
...
+        } else {
+            botSays.innerHTML = `Sorry, we don't serve ${answer}. Please order something from the menu`;
+        }
+
+        document.getElementById("answer").value = "";
+    } else {
+        // GET ORDER QUANTITY
+        let orderQuantity = answer
+
+        // CALCULATE COST
```
EmOlsz, line 66: These are nice comments, they help me (and whoever it the future will analyse this code) a lot, good job :)

code/pizzabot.js
```diff
@@ -1,7 +1,89 @@
...
+        // Reset order mode for next customer
+        document.getElementById("answer").value = "";
+        orderedPizza = "";
+        document.getElementById("answer").placeholder = `enter pizza name`
+    }
+}
+
+let input = document.getElementById("answer");
+input.addEventListener("keyup", function (event) {
+    if (event.key === "Enter") {
```
EmOlsz, line 84: It's more secure to detect exact key by property `keyCode`. If I'm not wrong, "Enter' has `keyCode = 13`, so we could write this statement like this :
`if (event.keyCode === 13) { ... }`

code/style.css
```diff
@@ -0,0 +1,53 @@
...
+}
+
+body {
+	background-color: #f0db4f;
+}
+
+.container {
+    display: block;
+    padding-top: 100px;
+
```
EmOlsz: Unnecessary empty line over here

code/style.css
```diff
@@ -0,0 +1,53 @@
...
+	color: #323330;
+	font-size:  100px;
+}
+
+
+p {
+	font-size: 30px;
+}
+
+#output, #container {
```
EmOlsz, line 26: We shouldn't style by id's, let's change them to class instead

EmOlsz (approved):
Great job Sara & Liza, I'm impressed :) It's especially because of the way your applications works - very intuitive and user friendly, keep it like this! Your code is also pretty clean, maybe apart from some comments you can find below. What's most important for me though is that it's logical and consequent, I believe that's also the most important thing from Week 4 to learn. In my opinion you achieved that :)

Before I'll be able to officially approve your project, you need to fix some stuff in your code :
1. Change `let` to `const` where needed
2. Decided whether you want to use semicolons or not and change your code

EmOlsz (changes requested):
Great job Sara & Liza, I'm impressed :) It's especially because of the way your applications works - very intuitive and user friendly, keep it like this! Your code is also pretty clean, maybe apart from some comments you can find below. What's most important for me though is that it's logical and consequent, I believe that's also the most important thing from Week 4 to learn. In my opinion you achieved that :)

Before I'll be able to officially approve your project, you need to fix some stuff in your code :
1. Change `let` to `const` where needed
2. Decided whether you want to use semicolons or not and change your code
3. Delete unnecessary empty line of code from CSS file

EmOlsz (approved):

PR #47, opened 2020-09-12
Project 4: Pizza Bot - Jamie Cook & Andrea Osmar

the-j-curl:
Link to my deployed project: https://pizza-corner.netlify.app/

I feel disappointed that we were not able to complete more of the red level goals, and feel that our project is very basic in design, especially compared to some of our team mates pizza bots.
But we did not feel confident or comfortable making the next step up. We both understood how to create functions, but found it tough creating multiple functions and knowing when and how to call the next part of the process.
However I do feel proud that we were able to make the correct pizza images display at the end along with the confirmation text.

- How to structure multiple functions

-How to change all of the browser prompts to prompts inside the website

- In our javascript file (pizzabot.js) on line 13 we define a function `const askForPizza = (event) =>` to pause the initial alert box from showing until the button is pressed. Which works. But we also define the same function on line 31 ` const askForPizza = () =>`. When I try to move the code from line 13 to 31 so all the code is under the same function, the code no longer works and nothing happen when you click the button. I'm not sure why this is? And I'm not sure why having `const askForPizza =` twice doesn't cause problems?

- On line 42 (pizzabot.js) we have created a function called pizzaImgFunction to define which image to show at the end of the process. Firstly I'm not sure if this is the best way to do this. And secondly I tried to include the variable names of the pizza's (vegetarian, hawaiian and pepperoni) in the condition but this did not work. So I had to use a string instead, which does not feel as "clean". Example: `if(order === "vegetarian pizza")`


code/pizzabot.js
```diff
@@ -1,7 +1,69 @@
...
-//Put your Javscript code here:
+let orderQuantity;
+let orderTotal;
+let orderName;
+let cookingTime;
+let imageName;
+
+  const askForPizza = (event) => {
+    event.preventDefault();
+    alert(`Welcome to Pizza Corner! On our menu we have ${vegetarian}, ${hawaiian} and ${pepperoni}`);
```
HIPPIEKICK: This first function misses a closing curly bracket (you shouldn't be able to declare the same constant twice). I suggest you take a look at [this stack overflow question](https://stackoverflow.com/c/technigo/questions/1447/1475#1475) to get some guidance in how to loop this question.

code/pizzabot.js
```diff
@@ -1,7 +1,69 @@
...
+    if (
+      order === "vegetarian pizza" ||
+      order === "pepperoni pizza" ||
+      order === "hawaiian pizza" 
+    ) {
+      orderQuantity = prompt(`How many ${orderName}'s would you like?`);
+      orderQuantity = parseInt(orderQuantity, 10);
+    } else {
+      alert("Please choose a pizza from our menu");
+      askForPizza();
```
HIPPIEKICK: We should declare our functions before invoking them. See the stack overflow question in the comment above for inspiration

HIPPIEKICK: Here's a starting example for only one `askForPizza()` function, using variable names in the conditionals and not invoking the functions before declaration: https://codepen.io/hippiekick/pen/RwaYXOG

code/pizzabot.js
```diff
@@ -1,7 +1,69 @@
...
+    }
+  };
+
+  const askForPizza = () => {
+    //alert(`Welcome to Pizza Corner! On our menu we have ${vegetarian}, ${hawaiian} and ${pepperoni}`); 
+    orderName = prompt("Enter the name of the pizza you want to order today"); 
+    validateOrderName(orderName);
+  };
+
+  askForPizza();
```
HIPPIEKICK: It should only be invoked after click, right?

code/pizzabot.js
```diff
@@ -1,7 +1,69 @@
...
+
+  const pizzaImgFunction = (order) => {
+    if(order === "vegetarian pizza") {
+        return "./images/vegetarian-pizza.jpg";
+    } else if (order === "pepperoni pizza") {
+        return "./images/pepperoni-pizza.jpg";
+    } else {
+        return "./images/hawaiian-pizza.jpg";
+    }
+  };
```
HIPPIEKICK: I can't see a reason for not using the variables. Is it perhaps that you wrote lowercase letters and your variables contain uppercase letters? If you want to make it case insensitive, take a look at this question on stack overflow: https://stackoverflow.com/c/technigo/questions/1427

HIPPIEKICK: Naming functions is really hard, but we should strive to describe what the function is doing. Maybe this could be called something like `findPizzaImage` or `getPizzaImage`?

code/pizzabot.js
```diff
@@ -1,7 +1,69 @@
...
+      cookingTime = 15;
+    } else {
+      cookingTime = 20;
+    }
+    alert(`Great I'll get started on your ${orderName} order right away. It will cost ${orderTotal}kr. The pizza(s) will take ${cookingTime} minutes`);
+
+    document.getElementById("printOrder").innerHTML = `You have ordered: ${orderName} x${orderQuantity} and the total price is: ${orderTotal}kr`;
+    document.getElementById("pizzaImage").innerHTML = pizzaImage.src = pizzaImgFunction(orderName);
+
+  };
```
HIPPIEKICK: This function does a little bit more than calculate the cooking time. Consider renaming the function or dividing it to two functions where the second function can be called something like `finalizeOrder` or `displayFinalOrder`

code/index.html
```diff
@@ -4,9 +4,43 @@
...
+        <img src="./images/header-img-large.jpg" alt="Image of pizza - Photo by Brett Jordan from Pexels" />
+      </picture>
+      <button
+        id="orderBtn"
+        class="order-btn"
+        type="submit"
+        onclick="askForPizza(event)"
+      >
+        Order Pizza</button>
+    </header>
```
HIPPIEKICK: Super nice with a hero image! Not sure the button is a submit button though

code/index.html
```diff
@@ -4,9 +4,43 @@
...
+
+    <div class="pizza-order">
+      <p>Your order will display here once completed</p>
+      <p id="printOrder" class="print-order"></p>
```
HIPPIEKICK: To get rid of the first paragraph once the user has finished the order, you could make this to one paragraph and just change the content with JS

code/style.css
```diff
@@ -0,0 +1,137 @@
...
+  box-shadow: ;
+}
+
+.orderPizza {
+  display: flex;
+  color: green;
+  font-weight: 100;
+  font-size: 20px;
+  z-index: 1;
+}
```
HIPPIEKICK: You have no class with this name 🤔

code/style.css
```diff
@@ -0,0 +1,137 @@
...
+.print-order {
+  text-align: center;
+  color: black;
+  font-size: 20px;
+  margin: 20px;
+}
+
+.print-order {
+  font-weight: bold;
+}
```
HIPPIEKICK: Why two? :)

code/style.css
```diff
@@ -0,0 +1,137 @@
...
+  background-color: #f0e3dd;
+}
+
+.header {
+  width: 100%;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  height: 80vh;
```
HIPPIEKICK, line 15: This `height` is a little big in mobile, consider decreasing it in mobile and having `80vh` in tablet and up

code/style.css
```diff
@@ -0,0 +1,137 @@
...
+.pizza-image img {
+    width: 90%;
+}
+
+.form {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+}
```
HIPPIEKICK: This class doesn't exist either

code/index.html
```diff
@@ -4,9 +4,43 @@
...
+    <header class="header">
+      <div class="wrapper">
+        <h1 class="header-text">Pizza Corner</h1>
+        <h2 class="sub-header-text">Delicious Pizza 24/7</h2>
+      </div>
+      <picture>
+        <source srcset="./images/header-img-small.jpg" media="(max-width:600px)" />
+        <source srcset="./images/header-img-med.jpg" media="(max-width:900px)" />
+        <img src="./images/header-img-large.jpg" alt="Image of pizza - Photo by Brett Jordan from Pexels" />
+      </picture>
```
HIPPIEKICK: When we have two or more attributes, we should give them new lines for better readability:
```
<picture>
  <source
    srcset="./images/header-img-small.jpg"
    media="(max-width:600px)" />
  <source
    srcset="./images/header-img-med.jpg"
    media="(max-width:900px)" />
  <img
    src="./images/header-img-large.jpg"
    alt="Image of pizza - Photo by Brett Jordan from Pexels" />
</picture>
```

HIPPIEKICK (changes requested):
Great job you two! I think you should be proud instead of disappointed - it looks really good and you have some great reflection in your pull request comment. I think your design is clean and since this is the first JavaScript project I think it makes sense to focus on the JavaScript basics and not rushing through the steps just to get to the next level. If you - in the future - would like to go back to this project and try out more things, that's totally fine. It's a really nice feature with the button and the images at the end 💯 

For now, there are some small changes that need to be made:
- You should use the variables that you've created whenever possible (I made a comment about `toLowerCase(`) that you can try out but please reach out if there's something else that's causing a problem with the variables).
- Take away the duplicate `askForPizza()` and make sure to declare the functions before invoking them

I will approve after these two bullets are dealt with. If you have time, you can also take a look at this:
- You seem to have a good understanding of when to use let vs const, but take a look at the suggestions I gave you on the names of some functions.
- Remove unused code (CSS)

Again, you seem to have a good overall understanding so be proud! 💪

code/pizzabot.js
the-j-curl: Hi Matilda,
Thanks for your feedback, it has been useful. We have cleaned up the code so hopefully everything is okay now.
We do feel that this makes more sense now after just completing another JavaScript project.

HIPPIEKICK (approved):

PR #48, opened 2020-09-12
KarSan's pizzeria

MindstormingAB:
This code is the result of a pair programming collaboration between Sandrine Elander and Karan Mann.

Link to our deployed project: https://bigbotman.netlify.app/

Starting using JavaScript was a challenge and it felt good to share this experience with another student. The main obstacle was to find enough time for pair programming and to stick to the driver/navigator rules when the task got complicated.

Things we'd like to have clarified or explained in more detail: how to get rid of the zooming that automatically enlarge form inputs on iPhone? See https://stackoverflow.com/c/technigo/questions/1471.


code/index.html
```diff
@@ -4,9 +4,74 @@
...
+
+      <div class="bubble-container" id="enterName-container">
+        <img src="./assets/robot-solid.svg" alt="PizzaBot" class="bot-img">
+        <div class="bubble" id="enterName"></div>
+      </div>
+
+      <div class="bubble-answer-container" id="nameForm-container">
+        <form class="bubble-answer" id="nameForm" onsubmit="submitName(event)">
+          <input id="name" name="name" type="text" placeholder="Enter your name" required>
+          <input id="nameSubmit" class="submit" name="nameSubmit" type="submit" value="Submit">
```
EmOlsz: No need to add `name` attribute here, `<input type="submit" />` is a special part of each form, which is only responsible for submitting form. When you give `name` attribute to it, you are somehow saying "I want to store value from that field and send it to back-end". But there is no value important for back-end here, right?

code/index.html
```diff
@@ -4,9 +4,74 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
-    <title>Project 4: Pizzabot</title>
+    <link rel="stylesheet" href="./style.css">
```
EmOlsz: Happy to see `./` in path to file :)

code/pizzabot.js
```diff
@@ -4,4 +4,73 @@ const pepperoni = "Pepperoni Pizza"
 
 const pizzaPrice = 80
 
-//Put your Javscript code here:
+
+// validateOrderName() which should take the orderName variable as an argument and return true or false if the pizza is on the menu or not.
+// const validateOrderName = (orderName) => {
+//   if ((orderName === hawaiian) || (orderName === vegetarian) || (orderName === pepperoni)) {
```
EmOlsz: If it's not used code, it should me removed before pushing your code to github

code/pizzabot.js
```diff
@@ -4,4 +4,73 @@ const pepperoni = "Pepperoni Pizza"
...
+//     return true;
+//   } else {
+//     return false;
+//   }
+// }
+
+// calculateTotalCost() which takes orderQuantity and pizzaPrice as arguments and returns the total cost for the order.
```
EmOlsz, line 23: Very descriptive comments, thank you :)

code/pizzabot.js
```diff
@@ -4,4 +4,73 @@ const pepperoni = "Pepperoni Pizza"
...
+const calculationTotalCost = (orderQuantity, pizzaPrice) => {
+  return orderQuantity * pizzaPrice;
+}
```
EmOlsz: Very nice :) Another tip from my side - since we have only one line inside `calculationTotalCost` function, we can write it in "arrow function shorthand" :

`const calculationTotalCost = (orderQuantity, pizzaPrice) => orderQty * pizzaPrice;`

code/pizzabot.js
```diff
@@ -4,4 +4,73 @@ const pepperoni = "Pepperoni Pizza"
...
+
+// calculateCookingTime() which takes orderQuantity and returns the number of minutes it will take to finish the order.
+const calculateCookingTime = (orderQuantity) => {
+  if (orderQuantity <= 2) {
+    return 10;
+  } else if (orderQuantity > 2 && orderQuantity <= 5) {
+    return 15;
+  } else {
+    return 20;
```
EmOlsz, line 33: Great job :)
Apart from returning given thing from function, `return` keyword does one more thing - it automatically finishes function. That's why we mustn't write anything below `return`, because that code will never be executed (of course in your case we have code below return true, but it's inside if statement, so it wouldn't enter second statement in the first place).
```
const calculateCookingTime = (orderQuantity) => {
  if (orderQuantity <= 2) {
    return 10;
  } else if (orderQuantity > 2 && orderQuantity <= 5) {
    return 15;
  }
  return 20;
}
```
As you can see, we are not writing `else` statement anymore. We can do that thanks to behaviour mentioned above; if `orderQuantity` is lower than 6, we will execute return true code from line 25 or 27 and finish function. If, on the other hand, `orderQuantity` is higher or equal 6, we will never enter line 25 or 27 and jump straight into line 29. After executing line 29 we will finish function same way :)

code/pizzabot.js
```diff
@@ -4,4 +4,73 @@ const pepperoni = "Pepperoni Pizza"
...
+  }
+}
+
+const greeting = (`Hey! Happy to serve your pizza. On our menu we have ${vegetarian}, ${hawaiian} and ${pepperoni}.`);
+document.getElementById('greeting').innerHTML = greeting;
+
+const greet = (event) => {
```
EmOlsz, line 40: There is an arrow function convention which says that if we pass to our function only one argument, we can omit parentheses `()`, and rather write :

`const greet = event => { ... }`

than

`const greet = (event) => { ... }`

code/pizzabot.js
```diff
@@ -4,4 +4,73 @@ const pepperoni = "Pepperoni Pizza"
...
+  event.preventDefault();
+  const enterName = ('What is your name?');
```
EmOlsz: No need for parentheses over here :)

code/pizzabot.js
```diff
@@ -4,4 +4,73 @@ const pepperoni = "Pepperoni Pizza"
...
+  document.getElementById('enterName-container').style.display = "flex";
+  document.getElementById('enterName').innerHTML = enterName;
+  document.getElementById('nameForm-container').style.display = "flex";
+  document.getElementById('nameForm').style.display = "flex";
+}
+
+const submitName = (event) => {
+  event.preventDefault();
+  const name = document.getElementById('name').value;
+  const orderName = (`${name}, enter the name of the pizza you want to order today.`);
```
EmOlsz: Hmm, that value could be stored in global variable - there is a lot of code repetition.
Same goes for lines below in each function - as you can see, you are doing same set of things many times, for each step. Couldn't you universalised you code flow so there would be only one function for each step?

code/style.css
```diff
@@ -0,0 +1,153 @@
+body {
+  background: #e0e1dd;
+  color: #393d3f;
+  font-family: 'Athiti', Arial, Helvetica, sans-serif;
```
EmOlsz, line 4: Happy to see font fallback!

code/style.css
```diff
@@ -0,0 +1,153 @@
...
+  border-style: inset;
+  margin-bottom: 10px;
+}
+
+.submit {
+  width: 70px;
+  margin-bottom: 0;
+}
+
+#nameForm {
```
EmOlsz: We shouldn't style by id's, it's always better to add class to that element as well and reference to it in CSS

code/style.css
```diff
@@ -0,0 +1,153 @@
...
+  display: none;
+}
+
+#quantityForm {
+  display: none;
+}
+
+/* MEDIA QUERIES */
+
+@media (min-width: 768px) {
```
EmOlsz, line 104: Nice :)

EmOlsz (changes requested):
Well done Karan and Sandrine, you made good project this week :) I'm happy to see nice HTML visualisation and many different form elements! Your JS flow is proper, however it could be universalised. Apart from that, I can see you understand JavaScript concepts and implementation and I consider it as the most important thing to learn in Week 4. Good job :)

Before I'll be able to officially approve your project, I need you to add some changes. Those are :
1. Remove `name` attribute from `<input type="submit" />`
2. Remove commented code
3. Remove parentheses from string variables declarations
4. I've noticed there is no validation in orderQuantity field, I was able to order -2 pizzas ;) let's fix that as well

code/pizzabot.js
MindstormingAB: I created some global variables but, right now, I don't really see how to minimize the code flow in order to keep only one function for each step...

MindstormingAB:
Hi @EmOlsz 
The four changes you required have been done and a new version has been committed.

code/pizzabot.js
EmOlsz: You could for example create function :

```
const addFlexStyle  = elementId => {
    document.getElementById(elementId).style.display = "flex";
}

addFlexStyle('pizzaForm-container');
```

Something like that :)

Nevertheless, project is officially approved!

EmOlsz (approved):

PR #49, opened 2020-09-13
Pizza Bot - Johanna Rexin and Amanda Flodström

amandaflod:
Here is our pizza bot project. Have a nice Sunday!


code/index.html
```diff
@@ -1,12 +1,54 @@
...
-    <h1>This is your Pizza Bot</h1>
-  </body>
-</html>
+
+<head>
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+  <title>Project 4: Pizzabot</title>
+  <link rel="stylesheet" href="style.css">
```

code/index.html
```diff
@@ -1,12 +1,54 @@
...
+  <link href="https://fonts.googleapis.com/css2?family=Pacifico&family=Special+Elite&display=swap" rel="stylesheet">
+</head>
+
+<body>
+
+  <header>
+    <img src="pizza.jpg">
```
EmOlsz: `alt` attribute missing

code/index.html
```diff
@@ -1,12 +1,54 @@
...
+      <label class="formItemLabel" for="pizzas">Choose pizza type:</label>
+      <select id="pizzas">
+        <option value="Vegetarian Pizza">Vegetarian</option>
+        <option value="Hawaiian Pizza">Hawaiian</option>
+        <option value="Pepperoni Pizza">Pepperoni</option>
+        <option value="Margherita Pizza">Margherita</option>
+      </select>
+      <label class="formItemLabel" for="number">How many pizzas do you want?</label>
+      <input class="formItemInput" type="number" id="numberOfPizzas" min="0">
+      <input class="formItem" type="button" id="submit" value="Submit">
```
EmOlsz, line 34: Your button should have `type="submit"` rather than `type="button"`

code/pizzabot.js
```diff
@@ -1,7 +1,65 @@
...
 const pizzaPrice = 80
 
-//Put your Javscript code here:
+const pizzaVegImg = document.getElementById("veg")
+const pizzaHawImg = document.getElementById("haw")
+const pizzaPepImg = document.getElementById("pep")
+const pizzaMarImg = document.getElementById("mar")
+
+
+console.log(vegetarian, hawaiian, pepperoni, pizzaPrice)
```
EmOlsz: Before pushing your code to github, always check if there are any console.log() and delete them - it's considered bad practise to push them. Everyone can see your code out here and it's pretty messy

code/pizzabot.js
```diff
@@ -1,7 +1,65 @@
...
+
+document.getElementById("menu").innerHTML = `Hey! Happy to serve your pizza. On our menu we have ${vegetarian}, ${hawaiian}, ${pepperoni}, and ${margherita}.`
+
+
+const calculateTotalCost = (orderQuantity) => {
+  return orderQuantity * pizzaPrice
+}
```
EmOlsz: Very nice :) Since we have only one line inside `calculateTotalCost `function, we can write it in "arrow function shorthand" :

`const calculateTotalCost = orderQty => orderQty * pizzaPrice;`

code/pizzabot.js
```diff
@@ -1,7 +1,65 @@
...
+
+document.getElementById("submit").addEventListener("click", handleClick)
+
+
+
+
+
+
+
+
```
EmOlsz: Why so many empty lines?

code/pizzabot.js
```diff
@@ -1,7 +1,65 @@
...
+const pizzaPepImg = document.getElementById("pep")
+const pizzaMarImg = document.getElementById("mar")
+
+
+console.log(vegetarian, hawaiian, pepperoni, pizzaPrice)
+
+document.getElementById("menu").innerHTML = `Hey! Happy to serve your pizza. On our menu we have ${vegetarian}, ${hawaiian}, ${pepperoni}, and ${margherita}.`
+
+
+const calculateTotalCost = (orderQuantity) => {
```

`const calculateTotalCost = orderQuantity => { ... }`

than

`const calculateTotalCost = orderQuantity => { ... }`

code/pizzabot.js
```diff
@@ -1,7 +1,65 @@
...
+  return orderQuantity * pizzaPrice
+}
+
+const calculateCookingTime = (orderQuantity) => {
```
EmOlsz: That function could be done in a better way. First of all, you don't actually need this local variables `cooking time`. Take a look how we could write same function in a more clean way :
```
const calculateCookingTime = orderQuantity => {
  if (orderQuantity <= 2) {
    return "10 minutes"
  } else if (orderQuantity >= 3 && orderQuantity <= 5) {
    return "15 minutes"
  } else {
    return "20 minutes"
  }
}
```
Apart from returning given thing from function, `return` keyword does one more thing - it automatically finishes function. That's why we mustn't write anything below `return`, because that code will never be executed (of course in your case we have code below `return`, but it's inside if statement, so it wouldn't enter second statement in the first place).
```
const calculateCookingTime = orderQuantity => {
  if (orderQuantity <= 2) {
    return "10 minutes"
  } else if (orderQuantity >= 3 && orderQuantity <= 5) {
    return "15 minutes"
  }
  return "20 minutes"
}
```
As you can see, we are not writing `else` statement anymore. We can do that thanks to behaviour mentioned above; if result of if statement is lower than 6, we will execute `return "10 minutes"` or `return "15 minutes"` code and finish function. If, on the other hand, result of it was higher or equal 6, we would never enter these two statements and jump straight into line`return "20 minutes"`. After executing that line, we would finish function same way :)

code/style.css
```diff
@@ -0,0 +1,144 @@
...
+}
+
+header img{
+  width: 100%;
+  height: 100vh;
+  background-size: cover;
+  background-position: center center;
+  background-repeat: no-repeat;
+  background-attachment: fixed;
+  filter: brightness(70%);
```
EmOlsz, line 23: Nice :)

code/style.css
```diff
@@ -0,0 +1,144 @@
...
+header h1{
+  text-align: center;
+  position: absolute;
+  font-family: 'Pacifico', cursive;
+  color: #F3F0DD;
+  text-shadow: 0px 0px 15px black;
+  font-size: 5rem;
+}
+
+#orderMenu {
```
EmOlsz: You seem to style a lot by ids and selectors - please, change that to classes only. Styling by ids and selectors is considered as bad practice because styles added by ids have higher importance than by class, and styles added by selectors have lower importance than by class

EmOlsz (changes requested):
Brawo Johanna and Amanda, nice project this week :) I've added some comments below, please check them out. In general, you JS flow and logic is proper, you made it in a nice way, and I'm happy to see HTML visualisation!

Before I'll be able to officially approve your project, I need you to add some changes. Those are :

1. Add `alt` attribute to images
2. Remove `console.log()`
3. Remove empty lines of code
4. Change id and selector references to class references in CSS

amandaflod:
Hey Maks!

Thank you for your comments, we have now updated the code and pushed the changes. Is it correct that we do not need to do another pull request?

Have a great weekend!

EmOlsz (approved):

EmOlsz:
Hey Johanna and Amanda, project approved :) No need for another pull request, I can see your changes

